PR #1, opened 2017-12-11
Update requirements and move backup-pruner to group vars.

UmanShahzad:
This PR makes the following fixes:

1. Moves the backup-pruner variables that were in `host_vars` to `group_vars`.
1. Removes unnecessary `cryptography==1.3.1`.
1. Updates the `ansible-role-security` to use the latest upstream master -- I checked the commit history since the commit we were using, and it is mostly just doc/test fixes & updates, and https://github.com/geerlingguy/ansible-role-security/commit/2abe44e34386f66633dd512d8a286e413c232d3f which adds the optional `security_autoupdate_blacklist`.
1. Updates the `ansible-tarsnap` to use a commit hash rather than a branch since it's an external role.

### DIFF
--- a/group_vars/backup-pruner/public.yml
+++ b/group_vars/backup-pruner/public.yml
@@ -1,6 +1,3 @@
-FORWARD_MAIL_RELAY_MAIL_TO: "ops@example.com"
-FORWARD_MAIL_SMTP_FROM: "{{ FORWARD_MAIL_RELAY_MAIL_TO }}"
-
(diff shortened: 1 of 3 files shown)


smarnach:
@UmanShahzad This looks good to me and fixes the points I mentioned on the ticket.  I'll just merge this without further testing, since we still need to test the whole repo anyway, so there isn't much point in testing this separately.

PR #2, opened 2017-12-13
Remove extraneous variable override, and add bootstrap template

kaizoku:
This PR contains a fix where the backup directory was erroneously overrode, causing disks to fill early.

It also moves a playbook into the public repo as it did not contain any secrets.


smarnach:
:+1: Looks good to me.  See the notes on https://github.com/open-craft/deployment-deploy-databases/pull/11 as well.

The bootstrap playbook does not need any additional testing, since it is a standalone playbook that was simply moved here from ansible-secrets.

PR #5, opened 2018-01-30
Dedicated servers should not set up a swapfile.

xitij2000:
When running on dedicated servers, a swapfile should not be created. 

**Testing instructions**:
Testing this pull request requires setting up an unused server and running the dedicated server playbook on it. After running the playbook there should not be any swapfile set up on the server.

**Reviewers**
- [ ] @smarnach

### DIFF
--- a/group_vars/all/public.yml
+++ b/group_vars/all/public.yml
@@ -11,3 +11,6 @@ security_autoupdate_mail_to: '{{ OPS_EMAIL }}'
 
 # Certbot
 certbot_auto_renew_user: root
+
+# swapfile
+swapfile_size: "{{ '512MB' if (dedicated_server is undefined or not dedicated_server) else false }}"


group_vars/all/public.yml
```diff
@@ -11,3 +11,6 @@ security_autoupdate_mail_to: '{{ OPS_EMAIL }}'
 
 # Certbot
 certbot_auto_renew_user: root
+
+# swapfile
+swapfile_size: false if (dedicated_server is undefined or not dedicated_server) else '512MB'
```
smarnach: Doesn't this need to go inside `"{{  }}"` to be processed by Jinja2?  I'd be surprised if it works like this.

xitij2000: Yup, sorry, I think I forgot to push the latest amend.

smarnach:
@xitij2000 I just tested this, and it appears the condition is the wrong way around, i.e. it disables the swap files for all VMs.

xitij2000:
@smarnach Updated.

smarnach:
@xitij2000 Thanks!  I've retested this with a dry run for both a dedicated server and a VM, and it's working fine now.  As I said on the ticket, deploying this to production isn't going to have any effect, since the `swapfile` role does not delete existing swap files.  So we can simply merge the PRs here, and delete and disable the swap files manually on the dedicated servers.

PR #6, opened 2018-02-02
Add RabbitMQ node health check

clemente:
This adds a command that checks whether rabbitmq passes the health check. If the connection fails for any other reason, the output won't contain the desired message, the `grep` will fail and an error e-mail will be sent by [sanity-checker](https://github.com/open-craft/ansible-sanity-checker).

### DIFF
--- a/group_vars/rabbitmq/public.yml
+++ b/group_vars/rabbitmq/public.yml
@@ -6,5 +6,9 @@ SANITY_CHECK_LIVE_PORTS:
     port: 15671
     message: "RabbitMQ management does not respond on port 15671."
 
+SANITY_CHECK_COMMANDS:
+  - command: /usr/sbin/rabbitmqctl node_health_check
+    message: "RabbitMQ health check not passing"
+
 TARSNAP_BACKUP_FOLDERS: "{{ RABBITMQ_EXPORT_DIR }} /etc/rabbitmq /etc/letsencrypt"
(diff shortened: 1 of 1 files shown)


group_vars/rabbitmq/public.yml
```diff
@@ -6,5 +6,9 @@ SANITY_CHECK_LIVE_PORTS:
     port: 15671
     message: "RabbitMQ management does not respond on port 15671."
 
+SANITY_CHECK_COMMANDS:
+  - command: /usr/sbin/rabbitmqctl node_health_check | grep 'Health check passed'
```
smarnach: Any reason for the `grep`?  The return code of `rabbitmqctl` seems already correct.

clemente: The [documentation](https://www.rabbitmq.com/rabbitmqctl.8.html#node_health_check) doesn't specify the return code. The command could write some warning (like „some tests passed, others not“) and still return 0; probably not, but I'd need to check the source code to be sure, therefore better to use something we control.
I can change it if we trust it.

smarnach: Relying on the exact textual output seems a lot more fragile than relying on the return code to me.  I tried that the return code is non-zero when the memory alert is active, so at least the error condition we know would be covered.  So I'd ask you to remove the `grep` here.  Once you have done this, you can deploy this to `rabbitmq-stage`.

clemente: Being fragile is a feature for a health checker :-) But ok, I removed it and will merge it and deploy carefully to stage.

PR #7, opened 2018-02-06
Add new server type for MySQL slaves

kaizoku:
This adds configuration to the MySQL server to allow for replication, and adds a new group for MySQL slave servers.

### DIFF
--- a/deploy-all.yml
+++ b/deploy-all.yml
@@ -29,6 +29,15 @@
     - geerlingguy.mysql
     - forward-server-mail
 
+- name: Set up MySQL slave servers
+  hosts: mysql-slave
+  become: true
+  roles:
+    - common-server
+    - mysql
+    - geerlingguy.mysql
+    - forward-server-mail
+
 - name: Set up PostgreSQL servers
   hosts: postgres
   become: true
--- a/group_vars/mysql-slave/public.yml
+++ b/group_vars/mysql-slave/public.yml
@@ -1,21 +1,15 @@
 MYSQL_SERVER_DOMAIN: "{{ inventory_hostname }}"
 
-MYSQL_SERVER_BACKUP_DIR: '/var/lib/mysql/backup'
-TARSNAP_BACKUP_PRE_SCRIPT: "/usr/local/sbin/backup-pre.sh"
-TARSNAP_BACKUP_POST_SCRIPT: "/usr/local/sbin/backup-post.sh"
-TARSNAP_BACKUP_FOLDERS: "{{ MYSQL_SERVER_BACKUP_DIR }}"
-TARSNAP_ARCHIVE_NAME: "{{ MYSQL_SERVER_DOMAIN }}"
+COMMON_SERVER_NO_BACKUPS: True
 
 mysql_root_password_update: true
 
 mysql_packages:
-  - mysql-server-5.5
-  - mysql-client-5.5
+  - mysql-server-5.6
+  - mysql-client-5.6
 
-mysql_table_open_cache: 200000
-
-mysql_server_id: "1"
-mysql_replication_role: 'master'
+mysql_server_id: "{{ MYSQL_SERVER_ID }}"
+mysql_replication_role: 'slave'
 mysql_replication_master: "{{ MYSQL_REPLICATION_MASTER_HOST }}"
 mysql_replication_user:
     name: "{{ MYSQL_REPLICATION_USERNAME }}"
@@ -29,5 +23,3 @@ SANITY_CHECK_LIVE_PORTS:
   - host: localhost
     port: 3306
     message: "Can't connect to mysql"
-
-SANITY_CHECK_COMMANDS: []


group_vars/mysql-slave/public.yml
```diff
@@ -0,0 +1,31 @@
...
+TARSNAP_BACKUP_FOLDERS: "{{ MYSQL_SERVER_BACKUP_DIR }}"
+TARSNAP_ARCHIVE_NAME: "{{ MYSQL_SERVER_DOMAIN }}"
+
+mysql_root_password_update: true
+
+mysql_packages:
+  - mysql-server-5.6
+  - mysql-client-5.6
+
+mysql_server_id: "{{ MYSQL_SERVER_ID | default('99') }}"
```
smarnach: This ID needs to be unique across the replica set, right?  In that case, I suggest removing the default to avoid unintended conflicts.

group_vars/mysql-slave/public.yml
```diff
@@ -0,0 +1,31 @@
...
+mysql_replication_role: 'slave'
```
smarnach: So if we wanted to fail over to the slave, we would need to reconfigure it to be a master instead and restart it, right?

group_vars/mysql/public.yml
```diff
@@ -9,8 +9,16 @@ TARSNAP_ARCHIVE_NAME: "{{ MYSQL_SERVER_DOMAIN }}"
...
+  - mysql-server-5.6
+  - mysql-client-5.6
+
+mysql_server_id: "1"
+mysql_replication_role: 'master'
+mysql_replication_master: "{{ MYSQL_REPLICATION_MASTER_HOST }}"
+mysql_replication_user:
+    name: "{{ MYSQL_REPLICATION_USERNAME }}"
+    password: "{{ MYSQL_REPLICATION_PASSWORD }}"
+    host: "%"
```
smarnach: I had a quick look at the documentation of the mysql Ansible role we are using.  Do we also need to set `mysql_max_binlog_size`?  Apparently that's only the file size after which the binary log file gets rotated, and defaults to 1GB, but our role reduces it to 100MB.  Probably does not really matter, though.

deploy-all.yml
```diff
@@ -29,6 +29,15 @@
     - geerlingguy.mysql
     - forward-server-mail
 
+- name: Set up MySQL slave servers
+  hosts: mysql-slave
```
smarnach: Can't we simply add the new group `mysql-slave` to the play we already have?  It seems to be identical.

group_vars/mysql/public.yml
```diff
@@ -9,8 +9,16 @@ TARSNAP_ARCHIVE_NAME: "{{ MYSQL_SERVER_DOMAIN }}"
 mysql_root_password_update: true
 
 mysql_packages:
-  - mysql-server-5.5
-  - mysql-client-5.5
+  - mysql-server-5.6
```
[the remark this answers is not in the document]
kaizoku: Ah, you're right. `mysql-im.opencraft.hosting` had mysql-server-5.6, which is why I updated this here.
I can remove these from the PR, and just update them locally when performing the deployment.

group_vars/mysql/public.yml
```diff
@@ -9,8 +9,16 @@ TARSNAP_ARCHIVE_NAME: "{{ MYSQL_SERVER_DOMAIN }}"
...
+  - mysql-client-5.6
+
+mysql_server_id: "1"
+mysql_replication_role: 'master'
+mysql_replication_master: "{{ MYSQL_REPLICATION_MASTER_HOST }}"
```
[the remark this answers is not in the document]
kaizoku: For the master, geerlingguy.mysql [checks that the `mysql_replication_master` variable is not-null](https://github.com/geerlingguy/ansible-role-mysql/blob/47d626e34c14dbc6dee767e1343d9fd8714debd1/tasks/replication.yml#L12), but it doesn't really use the variable in the configuration. So.. it matters that it's set, but doesn't seem to matter what it's set to. I used the FQDN as that's the same as the slave.

group_vars/mysql-slave/public.yml
```diff
@@ -0,0 +1,31 @@
+MYSQL_SERVER_DOMAIN: "{{ inventory_hostname }}"
+
+MYSQL_SERVER_BACKUP_DIR: '/var/lib/mysql/backup'
```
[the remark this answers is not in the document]
kaizoku: I suppose we may not need backups unless we failover, and in that case we'll have to do some other reconfiguration of the slave to make it a master. We can configure backups at that point in time.
I'll replace these settings and the tarsnap settings with ```COMMON_SERVER_NO_BACKUPS: True```

deploy-all.yml
kaizoku: The `mysql_server_id` and `mysql_replication_role` are necessarily different, which is why I put this in a different role. I'm not so great with ansible though, so perhaps there's an easier way to do this and avoid duplication?

Also I've made some other changes to logging and such now, so the roles are more different.

group_vars/mysql-slave/public.yml
kaizoku: Yes it does. I'll remove the default.

group_vars/mysql/public.yml
kaizoku: We could set this back to 1GB if 100MB seems too low. I'm not sure how much difference it will make either way though.
There could be a small performance impact to frequent file rotations I suppose? Otherwise I'm not sure how different it will be.

PR #9, opened 2018-02-07
Add nginx to load balancer config to serve error pages

xitij2000:
This pull request adds the Nginx role to the load balancer to serve maintenance pages from the load balancer itself.

### DIFF
--- a/deploy-all.yml
+++ b/deploy-all.yml
@@ -73,6 +73,7 @@
   roles:
     - common-server
     - forward-server-mail
+    - geerlingguy.nginx
     - load-balancer
 
 - name: Set up rabbitmq servers


deploy-all.yml
```diff
@@ -73,6 +73,7 @@
   roles:
     - common-server
     - forward-server-mail
+    - geerlingguy.nginx
```
smarnach: I usually did not bother to use any nginx role, since it's so simple to make do without.  Installing nginx is a matter of

    - apt:
        name: nginx

After that, you just need to copy a config file and enable it.  If the role works for you, that's fine too, but if it's difficult to make it work I suggest falling back to the manual approach.

smarnach: Since you are relying on this role in {{ansible-load-balancer}}, you should add it to the dependencies of that role (in {{meta/main.yml}}, instead of adding it here.  (That is, if you decide that the role is useful.)

xitij2000: OK.

xitij2000: I'm taking the approach you recommended of installing Nginx using just the apt module. If that works out I will just close this PR>

PR #10, opened 2018-03-05
WIP for Prometheus prototype.

smarnach:


playbooks/roles/consul/defaults/main.yml
```diff
@@ -42,3 +42,6 @@ consul_performance_raft_multiplier: 5
 consul_service_config: ""
 
 consul_url: "https://releases.hashicorp.com/consul/{{ consul_version }}/consul_{{ consul_version }}_linux_amd64.zip"
+
+consul_http_public_port: null
```
UmanShahzad: Should we let the public port default to 18500 and then check for the password to decide to run the nginx proxy? We could then remove occurrences of `consul_http_public_port` in the host variables, and generally have a standard public port that can be easily changed in one place.

playbooks/roles/nginx-proxy/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+  apt:
+    name: python3-passlib
+  when: NGINX_PROXY_PASSWORD is not none
+
+- name: generate htpasswd file
+  htpasswd:
+    path: "{{ NGINX_PROXY_HTPASSWD_PATH }}"
+    username: "{{ NGINX_PROXY_USERNAME }}"
+    password: "{{ NGINX_PROXY_PASSWORD }}"
+  when: NGINX_PROXY_PASSWORD is not none
```
UmanShahzad: Should we make these two tasks a block?

playbooks/roles/nginx-proxy/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: set up nginx to reverse-proxy
+  template:
+    src: nginx-proxy.j2
+    dest: "{{ NGINX_PROXY_CONFIG_PATH }}"
+  notify:
+    - reload nginx
```
UmanShahzad: I haven't tested this to see if it works anyway but, should we add the handler for this in this role explicitly? It doesn't seem to exist in the files I see here.

playbooks/roles/prometheus/meta/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  - name: ernestas-poskus.prometheus
+    prometheus_node_exporter_install: false
+  - name: nginx-proxy
+    NGINX_PROXY_CONFIG_PATH: /etc/nginx/sites-available/prometheus
+    NGINX_PROXY_PUBLIC_PORT: 443
+    NGINX_PROXY_INTERNAL_PORT: 9090
+    NGINX_PROXY_USERNAME: "{{ PROMETHEUS_BASIC_AUTH_USERNAME }}"
+    NGINX_PROXY_PASSWORD: "{{ PROMETHEUS_BASIC_AUTH_PASSWORD }}"
+  - name: nginx-proxy
+    NGINX_PROXY_CONFIG_PATH: /etc/nginx/sites-available/grafana
```
UmanShahzad: I'm not sure how much the Grafana task will expand, but if it's going to expand quite a bit if we install stuff through Ansible for it, maybe it should be its own role?

playbooks/roles/prometheus/meta/main.yml
```diff
@@ -0,0 +1,16 @@
...
+    NGINX_PROXY_PUBLIC_PORT: 3443
+    NGINX_PROXY_INTERNAL_PORT: 3000
+    NGINX_PROXY_USERNAME: "{{ PROMETHEUS_BASIC_AUTH_USERNAME }}"
+    NGINX_PROXY_PASSWORD: "{{ PROMETHEUS_BASIC_AUTH_PASSWORD }}"
```
UmanShahzad: I do think we should just use the Grafana login screen here instead of having nginx basic auth, as you suggested from the ansible-secrets PR. If I understand correctly, we can configure the user then in the future through Grafana rather than having to run the nginx role here to update it.

playbooks/roles/prometheus/tasks/grafana.yml
```diff
@@ -0,0 +1,36 @@
...
+    regexp: "http_addr.*="
+    line: "http_addr = 127.0.0.1"
+
+- name: Grafana – start server
+  service:
+    name: grafana-server
+    state: started
+    enabled: yes
+
+# This task does not work on Python 3, and is commented out for this reason.
```
UmanShahzad: :(

I haven't looked into it much, but maybe we can use https://github.com/cloudalchemy/ansible-grafana/blob/master/tasks/datasources.yml ?

playbooks/appserver.yml
```diff
@@ -8,7 +8,7 @@
     - role: consul
       when: consul_servers | length > 0
     - role: node-exporter
-      when: NODE_EXPORTER_PASSWORD != 'set-me-please'
+      when: NODE_EXPORTER_PASSWORD is not none
```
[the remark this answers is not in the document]
smarnach: I've filed https://github.com/open-craft/opencraft/pull/312 to change this.

I don't think the integration tests would actually have failed, since the default on the Ocim side still was `set-me-please`.  However, the node exporter would still be installed even if the default isn't changed, and that is certainly undesired.

playbooks/roles/nginx-proxy/tasks/main.yml
smarnach: Roles can trigger handlers from other roles.  The "reload nginx" handler is defined in the common-server role.

playbooks/roles/consul/defaults/main.yml
smarnach: Sounds good as well.  My original reasoning was that someone might want to expose the API without setting a password, but we certainly don't, and it's probably not a good idea to even support it, since it's rather dangerous, so I made the change you proposed.

playbooks/roles/prometheus/meta/main.yml
smarnach: I'm not sure yet whether we will be able to completely configure it with Ansible.  If this turns out to be easy to do, I will do it, but otherwise I will simply manually configure it and configure a backup for the configuration, so we can restore it if we lose the server.

playbooks/roles/prometheus/meta/main.yml
smarnach: Yeah, in hidnsight this seems like the far better option.  Otherwise the password always has to be changed in sync, which is rather unexpected and confusing.

playbooks/roles/prometheus/tasks/grafana.yml
smarnach: @UmanShahzad I will look into this when setting up Grafana, which we have a separate ticket for.

I first need to figure out whether Python 2 is required on the target or on the host running Ansible.  If it's the former, we could use Python 2 on the target just for these tasks, which should be possible by adding

    vars:
      ansible_python_interpreter: /usr/bin/python

If it's the latter, we will have to keep using Python 2 virtualenvs for running Ansible, but this shouldn't really be a big deal – we just need to document it.

Alternatively, we may even be able to fix the bug and submit a patch upstream.

playbooks/appserver.yml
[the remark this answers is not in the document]
UmanShahzad: :+1:'d that PR

playbooks/roles/nginx-proxy/tasks/main.yml
UmanShahzad: Ah, ok :)

playbooks/roles/prometheus/meta/main.yml
UmanShahzad: Sounds good :+1:

playbooks/roles/prometheus/tasks/grafana.yml
UmanShahzad: Alright, let's leave this to the Grafana task.

PR #12, opened 2018-03-20
Add initial Consul configuration for server agents.

UmanShahzad:


group_vars/elasticsearch/public.yml
```diff
@@ -0,0 +1,23 @@
...
+elasticsearch_package: "elasticsearch=1.5.2"
+elasticsearch_network_host: 0.0.0.0
+elasticsearch_http_port: 9200
+elasticsearch_script_inline: false
+elasticsearch_script_indexed: false
+
+# CONSUL ######################################################################
+
+consul_service_config:
+  service:
```
smarnach: You've done a much more complete job than I intended for this ticket.  We actually only need the service definitions for Prometheus, and Prometheus will only monitor the exporters, not the services itself.  However, having service definitions for everything will enable lots of use cases we haven't planned yet, so it's nice to see this.

group_vars/load-balancer/public.yml
```diff
@@ -6,8 +14,38 @@ SANITY_CHECK_LIVE_PORTS:
...
-LOAD_BALANCER_SERVER_IP: "{{ COMMON_FLOATING_IP or ansible_host }}"
+# CONSUL ######################################################################
+
+consul_backup: true
+consul_service_config:
+  services:
+    - id: haproxy-http
+      name: "haproxy"
+      tags: ["haproxy"]
+      port: 80
```
smarnach: I think haproxy should advertise the floating IP instead of the main IP of the server.  Can we inlcude something like `addr: "{{COMMON_SERVER_IP }}"` here?  (I did not check that this is the right way of doing it, nor whether `addr` and `COMMON_SERVER_IP` are correct.)

smarnach: Of course this comment only applies to the dedicated server, not to the load balancers running on VMs, so maybe we don't need this here, but rather in the private repo.

requirements.yml
```diff
@@ -53,6 +53,10 @@
   src: https://github.com/open-craft/ansible-load-balancer
   version: origin/master
 
+- name: consul
+  src: https://github.com/open-craft/ansible-role-consul
+  version: uman/consul
```
smarnach: I always had to use `origin/uman/consul` in the past, but maybe the newer version of Ansible we are using now finally allows to omit the `origin/` part.  In any case, we will need to update this to the master branch before merging.

deploy-all.yml
```diff
@@ -27,6 +27,7 @@
     - common-server
     - mysql
     - geerlingguy.mysql
+    - consul
```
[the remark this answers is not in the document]
UmanShahzad: Yeah I had thought about this too, but that'll just add to the number of PRs and require more changes here, so for now I'd opt out of this if that's all good.

group_vars/elasticsearch/public.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+# ELASTICSEARCH ###############################################################
+
+elasticsearch_repo: "deb http://packages.elastic.co/elasticsearch/1.5/debian stable main"
+elasticsearch_package: "elasticsearch=1.5.2"
+elasticsearch_network_host: 0.0.0.0
+elasticsearch_http_port: 9200
+elasticsearch_script_inline: false
+elasticsearch_script_indexed: false
```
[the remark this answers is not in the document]
UmanShahzad: Yep, from the secrets repo. They aren't really secrets nor necessarily specific to the host they were in. I imagine we'd use them for any separate ElasticSearch instance we had to deploy for a client.

group_vars/elasticsearch/public.yml
```diff
@@ -0,0 +1,23 @@
...
+
+# CONSUL ######################################################################
+
+consul_service_config:
+  service:
+    name: "elasticsearch"
+    tags: ["elasticsearch"]
+    port: 9200
+    enable_tag_override: true
+    checks:
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

group_vars/elasticsearch/public.yml
UmanShahzad: Thanks a lot!

group_vars/load-balancer/public.yml
UmanShahzad: I believe the right key is `address` and the value `{{ COMMON_FLOATING_IP }}`.

Reference: https://www.consul.io/docs/agent/services.html

requirements.yml
UmanShahzad: Yeah somehow it managed to work, I also remember having to prepend `origin/`.

deploy-all.yml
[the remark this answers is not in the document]
smarnach: Fair enough, sounds good.

PR #13, opened 2018-03-26
Renew the certs after stopping nginx.

UmanShahzad:
Complements https://github.com/open-craft/ansible-mattermost/pull/3 but for the accounting service.


smarnach:
I can't really test this PR, since I don't have access to the accounting service, but the changes look good, so :+1:

PR #18, opened 2018-04-19
Disable Certbot installation for some groups.

smarnach:
The Certbot installation introduced in https://github.com/open-craft/ansible-common-server/pull/15 conflicts with the load-balancer role, so we need to disable it for the load balancers.  We also disable it for the Dalite server, since the corresponding role hasn't been fully migrated yet.

This PR should be merged after the following PRs:
* https://github.com/open-craft/ansible-common-server/pull/15
* https://github.com/open-craft/ansible-dalite/pull/6
* https://github.com/open-craft/ansible-load-balancer/pull/21
* https://github.com/open-craft/ansible-rabbitmq/pull/11
* https://github.com/open-craft/ansible-mattermost/pull/4
* https://github.com/open-craft/ansible-link-shortener/pull/2
* https://github.com/open-craft/ansible-accounting/pull/3
* https://gitlab.com/opencraft/dev/link-shortener/merge_requests/4

This PR also removes the Tarsnapper configuration file setting we had in place for all servers.  Tarsnapper is only needed on the backup-pruner, so this setting was some relic from the initial development of the Ansible infrastructure.

Before merging, the changes to `requirements.yml` need to be reverted.  After merging, the submodule hash in the secrets repo needs to be updated.

### DIFF
--- a/deploy-all.yml
+++ b/deploy-all.yml
@@ -1,5 +1,5 @@
 - name: Install Python 2
-  hosts: accounting:backup-pruner:chat:elasticsearch:load-balancer:mongodb:rabbitmq:relay:postgres
+  hosts: all
   become: true
   gather_facts: false
   tasks:
@@ -120,6 +120,4 @@
   roles:
     - common-server
     - forward-server-mail
-    - geerlingguy.nginx
-    - geerlingguy.certbot
     - accounting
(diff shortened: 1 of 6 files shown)


smarnach:
@UmanShahzad I'm merging this PR as well to make the follow-up PR less messy.  Any further comments can be addressed in the follow-up.

PR #19, opened 2018-04-20
Add Prometheus node exporter roles to requirements.

smarnach:
Before merging, the version of the common-server role needs to be reverted to `origin/master`.


UmanShahzad:
@smarnach I think you forgot to change the common-server role version before merging.

smarnach:
@UmanShahzad Yeah, I noticed – the node-explorer version as well.  Fixed in #20.

PR #22, opened 2018-04-22
Move node-exporter role to the end of the play for some servers.

smarnach:
As usual, we need to revert the changes to requirements.yml before merging, and update the submodule hash after merging.


UmanShahzad:
Looks good! Just waiting on a comment on JIRA before approval.

PR #24, opened 2018-04-25
Add filebeat role.

UmanShahzad:
This adds the initial version of the Filebeat role, afresh.

Note that the config here is a little complex. But in exchange, the actual configuration needed for individual groups of servers is very tiny. Otherwise the diff could be much larger.

### DIFF
--- a/playbooks/group_vars/backup-pruner/public.yml
+++ b/playbooks/group_vars/backup-pruner/public.yml
@@ -1,2 +1,14 @@
+---
+
+# COMMON ######################################################################
+
 COMMON_SERVER_NO_BACKUPS: true
 COMMON_SERVER_INSTALL_CONSUL: false
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "backup-pruner"
+    paths:
+      - "{{ BACKUP_PRUNER_LOG_DIR }}/*.log"
--- a/playbooks/group_vars/chat/public.yml
+++ b/playbooks/group_vars/chat/public.yml
@@ -8,3 +8,11 @@ SANITY_CHECK_LIVE_PORTS:
     message: "Cannot connect to HTTPS port."
 
 MATTERMOST_OPS_EMAIL: "{{ OPS_EMAIL }}"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mattermost"
+    paths:
+      - "{{ MATTERMOST_LOG_DIR }}/*.log"
--- a/playbooks/group_vars/dalite/public.yml
+++ b/playbooks/group_vars/dalite/public.yml
@@ -49,3 +49,11 @@ BACKUP_SWIFT_DEST: '/usr/local/sbin/backup-container.sh'
 BACKUP_SWIFT_CACHE: "/var/cache/tarsnap-swift"
 BACKUP_SWIFT_TARSNAP_KEY: "{{ DALITE_TARSNAP_SWIFT }}"
 BACKUP_SWIFT_RC: "{{ DALITE_SWIFT_BACKUP_RC }}"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "dalite"
+    paths:
+      - "{{ DALITE_LOG_DIR }}/*.log"
--- a/playbooks/group_vars/elasticsearch/public.yml
+++ b/playbooks/group_vars/elasticsearch/public.yml
@@ -21,3 +21,11 @@ consul_service_config:
       - name: "Connect to Elasticsearch"
         tcp: "{{ inventory_hostname }}:9200"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "elasticsearch"
+    paths:
+      - "{{ elasticsearch_log_dir }}/*.log"
--- a/playbooks/group_vars/mongodb/public.yml
+++ b/playbooks/group_vars/mongodb/public.yml
@@ -13,7 +13,7 @@ mongodb_storage_dbpath: "/var/lib/mongodb"
 # is using 14 GB for the module store.)
 mongodb_storage_engine: 'mmapv1'
 mongodb_storage_quota_maxfiles: 36
-mongodb_systemlog_path: "/var/lib/mongodb/log/{{ mongodb_daemon_name }}.log"
+mongodb_systemlog_path: "{{ mongodb_storage_dbpath }}/log/{{ mongodb_daemon_name }}.log"
 mongodb_pymongo_from_pip: true
 
 MONGODB_SERVER_DOMAIN: "{{ inventory_hostname }}"
@@ -53,3 +53,11 @@ consul_service_config:
       - name: "Connect to MongoDB"
         tcp: "{{ MONGODB_SERVER_DOMAIN }}:27017"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mongodb"
+    paths:
+      - "{{ mongodb_systemlog_path | dirname }}/*.log*"
--- a/playbooks/group_vars/mysql/public.yml
+++ b/playbooks/group_vars/mysql/public.yml
@@ -6,12 +6,10 @@ MYSQL_SERVER_DOMAIN: "{{ inventory_hostname }}"
 MYSQL_SERVER_BACKUP_DIR: '/var/lib/mysql/backup'
 mysql_root_password_update: true
 mysql_table_open_cache: 200000
+mysql_skip_name_resolve: true
 mysql_packages:
   - mysql-server-5.5
   - mysql-client-5.5
-
-mysql_table_open_cache: 200000
-mysql_skip_name_resolve: true
 
 # TARSNAP #####################################################################
 
@@ -45,3 +43,11 @@ consul_service_config:
       - name: "Connect to MySQL"
         tcp: "{{ MYSQL_SERVER_DOMAIN }}:3306"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mysql"
+    paths:
+      - /var/log/mysql.*
--- a/playbooks/roles/filebeat/tasks/main.yml
+++ b/playbooks/roles/filebeat/tasks/main.yml
@@ -1,0 +1,63 @@
+---
+
+- name: Install apt package dependencies.
+  apt:
+    name: apt-transport-https
+    state: present
+
+- name: Add Elastic's apt key
+  apt_key:
+    url: https://artifacts.elastic.co/GPG-KEY-elasticsearch
+    state: present
+
+- name: Add Elastic apt repository pin
+  template:
+    src: elastic.preferences.j2
+    dest: "/etc/apt/preferences.d/artifacts_elastic_co_packages_6_x_apt"
+
+- name: Add Elastic's apt repository
+  apt_repository:
+    repo: 'deb https://artifacts.elastic.co/packages/6.x/apt stable main'
+    state: present
+    update_cache: yes
+
+- name: Install Filebeat
+  apt:
+    name: filebeat
+    state: present
+
+- name: Install Filebeat configuration
+  template:
+    src: filebeat.yml.j2
+    dest: "{{ filebeat_config_dir }}/filebeat.yml"
+  notify:
+    - restart filebeat
+
+- name: Ensure Filebeat CA certificate variables are exclusive
+  assert:
+    that: filebeat_ca_cert == "" or filebeat_ca_cert_url == ""
+    msg: "Please set either `filebeat_ca_cert` or `filebeat_ca_cert_url`, but not both."
+
+- name: Install Filebeat CA certificate
+  copy:
+    content: "{{ filebeat_ca_cert }}"
+    dest: "{{ filebeat_ca_path }}"
+    owner: root
+    group: root
+    mode: 0400
+  when: filebeat_ca_cert != "" and filebeat_ca_cert_url == ""
+
+- name: Install Filebeat CA certificate from URL
+  get_url:
+    url: "{{ filebeat_ca_cert_url }}"
+    dest: "{{ filebeat_ca_path }}"
+    owner: root
+    group: root
+    mode: 0400
+  when: filebeat_ca_cert == "" and filebeat_ca_cert_url != ""
+
+- name: Ensure Filebeat is started and enabled at boot
+  service:
+    name: filebeat
+    state: started
+    enabled: yes
--- a/playbooks/roles/filebeat/templates/filebeat.yml.j2
+++ b/playbooks/roles/filebeat/templates/filebeat.yml.j2
@@ -1,0 +1,11 @@
+filebeat:
+  registry_file: "{{ filebeat_registry_file }}"
+
+  prospectors:
+    {% for prospector in filebeat_prospectors %}
+- {{ filebeat_common_prospector | combine(prospector, recursive=True) | to_nice_yaml | indent(width=6) }}
+    {% endfor %}
+
+output:
+  logstash:
+{{ filebeat_logstash | to_nice_yaml | indent(width=4, first=True) }}
--- a/playbooks/appserver.yml
+++ b/playbooks/appserver.yml
@@ -9,3 +9,7 @@
       when: consul_servers | length > 0
     - role: node-exporter
       when: NODE_EXPORTER_PASSWORD != 'set-me-please'
+    - role: filebeat
+      # Appservers may have some Elastic stuff already installed on them.
+      filebeat_elastic_apt_pin: 200
+      when: filebeat_logstash_hosts | length > 0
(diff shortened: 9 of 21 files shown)


UmanShahzad:
I need to rebase here on master. Will do that tomorrow.

playbooks/roles/filebeat/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    state: present
+
+- name: Add Elastic's apt key
+  apt_key:
+    url: https://artifacts.elastic.co/GPG-KEY-elasticsearch
+    state: present
+
+- name: Add Elastic's apt repository
+  apt_repository:
+    repo: 'deb https://artifacts.elastic.co/packages/6.x/apt stable main'
```
UmanShahzad: @smarnach Btw, I think this conflicts with the one upstream installs: https://github.com/edx/configuration/blob/master/playbooks/roles/elasticsearch/defaults/main.yml#L7

So if there is an upgrade run on the server it will probably upgrade Elasticsearch too if the service is local, is that right? Is this something to worry about?

playbooks/group_vars/mysql/public.yml
```diff
@@ -6,13 +6,11 @@ MYSQL_SERVER_DOMAIN: "{{ inventory_hostname }}"
 MYSQL_SERVER_BACKUP_DIR: '/var/lib/mysql/backup'
 mysql_root_password_update: true
 mysql_table_open_cache: 200000
+mysql_skip_name_resolve: true
 mysql_packages:
   - mysql-server-5.5
   - mysql-client-5.5
 
-mysql_table_open_cache: 200000
```
UmanShahzad: Got a warning that this already exists, and it does on line 8.

playbooks/roles/filebeat/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+
+- name: Install Filebeat
+  apt:
+    name: filebeat
+    state: present
+
+- name: Install Filebeat configuration
+  template:
+    src: filebeat.yml.j2
+    dest: "{{ filebeat_config_dir }}/filebeat.yml"
```
UmanShahzad: Note: we don't really need this var because it'll be `/etc/filebeat` by default through the apt package install, but since I already added it I can just leave it.

playbooks/roles/filebeat/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+- name: Install Filebeat CA certificate
+  copy:
+    content: "{{ filebeat_ca_cert }}"
+    dest: "{{ filebeat_ca_path }}"
+    owner: root
+    group: root
+    mode: 0400
+  when: filebeat_ca_cert != "" and filebeat_ca_cert_url == ""
+
+- name: Install Filebeat CA certificate from URL
```
UmanShahzad: There are some cases where adding the public cert through a var is pretty ugly or I think easy to mess up, like for Ocim appservers. I don't want to have to squeeze that public cert inside Ocim's `.env` somehow so that it can propgate to appservers running the playbook.

playbooks/roles/filebeat/templates/filebeat.yml.j2
```diff
@@ -0,0 +1,10 @@
+filebeat:
+  registry_file: "{{ filebeat_registry_file }}"
+
+  prospectors:
+    {% for prospector in filebeat_prospectors %}
+    - {{ filebeat_common_prospector | combine(prospector, recursive=True) }}
+    {% endfor %}
+
+output:
+  logstash: {{ filebeat_logstash }}
```
UmanShahzad: This file ends up looking kinda ugly, but I tried a lot of things to make it prettier. If anyone has ideas of how to make it clean while keeping it as valid YML and persisting the for loop above, do let me know.

playbooks/group_vars/mysql/public.yml
```diff
@@ -45,3 +43,11 @@ consul_service_config:
...
         tcp: "{{ MYSQL_SERVER_DOMAIN }}:3306"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mysql"
+    paths:
+      - /var/log/mysql.*
```
smarnach: It appears that the error logs in that directory are all empty, while the actual logs are in `/var/log/mysql.err`.  I don't fully understand why, though – seems to be some misconfiguration.

playbooks/roles/filebeat/meta/main.yml
```diff
@@ -0,0 +1 @@
+dependencies: []
```
smarnach: This file was only required for ansible-galaxy, so you can remove it now.

playbooks/roles/filebeat/tasks/main.yml
smarnach: Yeah, I didn't mean this as a strict policy.  It's more a habit I got into – making everything configurable just doesn't add much value, and in some case hurts readability, and is more work to set up, so I just only do it if I have to.

playbooks/roles/filebeat/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+  get_url:
+    url: "{{ filebeat_ca_cert_url }}"
+    dest: "{{ filebeat_ca_path }}"
+    owner: root
+    group: root
+    mode: 0400
+  when: filebeat_ca_cert == "" and filebeat_ca_cert_url != ""
```
smarnach: So if someone sets both `filebeat_ca_cert` and `filebeat_ca_cet_url`, none of them will be used?  Maybe add an `assert` to give an error message in that case – silently doing nothing instead seems somewhat unexpected.

playbooks/roles/filebeat/tasks/main.yml
UmanShahzad: :+1:

playbooks/accounting.yml
```diff
@@ -7,5 +7,5 @@
   become: true
   roles:
     - common-server
-    - forward-server-mail
     - accounting
+    - common-server-post
```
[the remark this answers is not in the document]
UmanShahzad: I at first thought Filebeat would cause issues if it found no logs as specified in the paths but it works just fine. So if we have nothing in particular to run in `-post`, we can delete that role. I'll need to rebase on your PR first though -- I'll wait for you to merge.

playbooks/group_vars/mysql/public.yml
UmanShahzad: Yeah, this is catching `/var/log/mysql.err`. There is definitely some misconfiguration.

playbooks/roles/filebeat/meta/main.yml
UmanShahzad: :+1:

playbooks/roles/filebeat/tasks/main.yml
UmanShahzad: Yeah didn't think about that. Will add an assert.

playbooks/group_vars/backup-pruner/public.yml
```diff
@@ -17,3 +17,11 @@ TARSNAPPER_JOBS:
...
     target: mysql-$date
     deltas: 1h 1d 7d 30d 360d 18000d
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "backup-pruner"
+    paths:
+      - /var/log/backup-pruner/*.log
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/group_vars/chat/public.yml
```diff
@@ -8,3 +8,11 @@ SANITY_CHECK_LIVE_PORTS:
...
 
 MATTERMOST_OPS_EMAIL: "{{ OPS_EMAIL }}"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mattermost"
+    paths:
+      - /var/log/mattermost/*.log
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/group_vars/dalite/public.yml
```diff
@@ -49,3 +49,11 @@ BACKUP_SWIFT_DEST: '/usr/local/sbin/backup-container.sh'
...
 BACKUP_SWIFT_TARSNAP_KEY: "{{ DALITE_TARSNAP_SWIFT }}"
 BACKUP_SWIFT_RC: "{{ DALITE_SWIFT_BACKUP_RC }}"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "dalite"
+    paths:
+      - /var/log/dalite/*.log
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/group_vars/elasticsearch/public.yml
```diff
@@ -21,3 +21,11 @@ consul_service_config:
...
         tcp: "{{ inventory_hostname }}:9200"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "elasticsearch"
+    paths:
+      - /var/log/elasticsearch/*.log
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/group_vars/mongodb/public.yml
```diff
@@ -53,3 +53,11 @@ consul_service_config:
...
         tcp: "{{ MONGODB_SERVER_DOMAIN }}:27017"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mongodb"
+    paths:
+      - /var/log/mongodb/*.log
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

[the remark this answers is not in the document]
UmanShahzad: This gives `mongod.log`, I suppose we also want `mongodb.log`? It seems quite spammy though.

playbooks/group_vars/mongodb/public.yml
```diff
@@ -53,3 +53,11 @@ consul_service_config:
...
         tcp: "{{ MONGODB_SERVER_DOMAIN }}:27017"
         interval: "10m"
+
+# FILEBEAT ####################################################################
+
+filebeat_prospectors:
+  - fields:
+      type: "mongodb"
+    paths:
+      - "{{ mongodb_systemlog_path }}"
```
smarnach: Is this file rotated by logrotate?  If so, do we need to add an asterisk at the end to also match the rotated files?

playbooks/group_vars/mysql/public.yml
smarnach: Sorry, I misread this.  And yeah, we can fix the misconfiguration in a different ticket, and it's not urgent.  As long as this doesn't cause any problems, we can ignore it as well.

playbooks/roles/filebeat/tasks/main.yml
smarnach: Yeah, we are running Elasticsearch mostly directly on the app servers, and we want to use version 1.5, because it's the only version that works.  If this repository adds a package with the name `elasticsearch` with a newer version than 1.5, it may be updated.  We could work around this by decreasing the priority of the repo we add in `/etc/apt/priorities` to a value that ensures that packages from that repo never override packages from other repos.

playbooks/roles/filebeat/tasks/main.yml
smarnach: An alternative would be to configure the URL in Ocim and make Ocim download the cert and render it in the settings, but the solution you have is fine, too.

playbooks/roles/filebeat/templates/filebeat.yml.j2
smarnach: Since this is a YAML file, I wouldn't define a template for it.  Just store the dictionary in an Ansible variable, and write it to a file using the `to_nice_yaml` filter.  This should render nicer YAML than you get with the current version.

UmanShahzad: The issue is that above I loop through the prospectors and merge them with the common prospector. I don't know if I can achieve something similar, of merging a dictionary with all dictionaries in a list, without a for loop, in plain YAML. Is there a way?

playbooks/group_vars/mysql/public.yml
UmanShahzad: :+1:

playbooks/group_vars/mongodb/public.yml
UmanShahzad: I don't see anything being rotated in that directory, so I don't think so. Actually, I don't know if I added a comment on this or not, but there's another, much larger file in that directory, `mongodb.log`.

```
1.7G	mongodb.log
8.0K	mongod.log
```

I suppose I should just define a new variable that leads to the log directory, and get `{{ log_dir }}/*.log`.

playbooks/roles/filebeat/tasks/main.yml
UmanShahzad: Yeah I just tested it, and it indeed got updated to 6.2.4. But it was the correct version of 1.5 after running all playbooks, so as long as no one "accidentally" runs an `apt-get upgrade` on any appserver manually, things should be alright. But we should definitely pin it anyway. I'll add a task for that.

UmanShahzad: After downgrading back to 1.5 and adding the priority (template in an upcoming commit), I get:

```
The following packages have been kept back:
  elasticsearch git git-man linux-headers-generic linux-headers-virtual linux-image-virtual linux-virtual rabbitmq-server
0 upgraded, 0 newly installed, 0 to remove and 8 not upgraded.
```

So it should be safe now.

UmanShahzad:
@smarnach I've addressed your comments and rebased. I'll take a look soon at the other PRs you have up because I'll need to rebase on them too once they get merged.

playbooks/roles/filebeat/templates/filebeat.yml.j2
smarnach: Ansible's "type system" (in quotes, since it doesn't really deserve the name) is rather arcane, so I can't foresee whether the result of the Jinja2 loop will be implicitly converted back to a YAML dictionary, or whether it's possible to explicitly convert it back.  I'm rather confident that this would be possible, e.g. using the `from_yaml` Jinja2 filter, but I don't care too much about the rendered template being ugly – your code looks fine, and this matters a lot more, so let's just not spend more time on this.

playbooks/group_vars/mongodb/public.yml
smarnach: Yeah, this looks rather broken.  I suggest simply going with `"{{ mongodb_systemlog_path | dirname }}/*.log*"` or similar for now, which should cover everything that's relevant.

smarnach:
@UmanShahzad The apt pinning you added looks good to me.  I'll re-review once this is fully rebased.

playbooks/roles/filebeat/templates/filebeat.yml.j2
UmanShahzad: :+1:

UmanShahzad:
@smarnach This is ready for another (hopefully final) review.

playbooks/roles/filebeat/templates/filebeat.yml.j2
```diff
@@ -0,0 +1,11 @@
+filebeat:
+  registry_file: "{{ filebeat_registry_file }}"
+
+  prospectors:
+    {% for prospector in filebeat_prospectors %}
+- {{ filebeat_common_prospector | combine(prospector, recursive=True) | to_nice_yaml | indent(width=6) }}
+    {% endfor %}
```
smarnach: I can see why it is behaving the way it behaves, but it's really not obvious.  This is a result of Jinja2's whitespace stripping behviour, which usuall behaves as intended, but can be difficult to exactly predict.  That's the reason why I suggested using `to_json` here – it will format the dictionary without any newline characters, so you don't run into the any problems with indentation.

For what it's worth, here is some code you could include in `defaults/main.yml` as an alternative:

    filebeat_prospectors_yaml: |
      {% for prospector in filebeat_prospectors %}
      - {{ filebeat_common_prospector | combine(prospector, recursive=True) | to_json }}
      {% endfor %}
    filebeat_config:
      filebeat:
        registry_file: "{{ filebeat_registry_file }}"
        prospectors: "{{ filebeat_prospectors_yaml | from_yaml }}"
      output:
        logstash: "{{ filebeat_logstash }}"

With these definitions, writing `filebeat.yml` becomes a matter of `{{ filebeat_config | to_nice_yaml }}`, and we don't need any fragile indentation code.  (I wrote this code just out of curiosity – completely up to you whether you want to include it.  It's still messy – we take a list of dictionaries, format them as JSON, combine them into a YAML-formatted string, then convert back to YAML just to format as YAML again.  The last two steps are for a nice-looking output.)

UmanShahzad: Yeah at this point I'll probably just leave this alone since I don't have more time to try other ways of doing this. I'd also like to keep the Jinja2 code inside of `j2` files to make it easier to understand what's going on. I think as long as one understands how `indent` is working in the current setup, things should be okay.

UmanShahzad:
@smarnach Thanks! Will merge it as is then.

PR #25, opened 2018-04-26
Upgrade Ansible to version 2.5.1, and clean the code up a bit.

smarnach:
The upgrade to 2.5.1 didn't require any major changes, and it allows us to use Python 3 on the target machines.  (We can now use either Python 2 or 3 on the machine running Ansible – it doesn't make a difference for the Python version used on the target machine.)  See the first commit for the changes that were required for the upgrade.

Apart from the upgrade, I also cleaned the code up a bit.
* I refactored the common-server role.  It now has a part called common-server-init, which contains all the tasks, and the original common-server role, which only contains dependencies.  This allows to run all the tasks before the dependencies, by including common-server-init as the first dependency.  This in turn allowed to include forward-server-mail as a dependency.  I tried to make this work using `import_role` instead.  However, when using `import_role` the variables of the role won't be persisted in the global variables, so I got errors e.g. when the tarsnap role tried to expand the list of directories that need to be backed up, which often contains variables from other roles (e.g. `consul_data_dir`).  The only solution I found was to split the role up, but it's actually not that bad.
* The configure-apt role was integrated into the common-server role.  It only was a separate role because it needed to run before the requirements, which is now solved by the common-server-init role.
* The cerbot was split out of the common-server role as a separate role.
* I used the multiline argument syntax for many more files.
* All unnecessary meta files have been removed.
* I made use of several newer Ansible features, e.g. passing a list of packages to the apt module, which significantly sped up the playbook runs.
* We don't run `tarnsap --fsck` and the sanity checker script anymore – the calls were slow, error-prone and unnecessary.
* Many tasks have been refactored in a way to better detect whether the target was changed, e.g. by adding `creates` arguments to shell commands.  The number of spurious "changed" tasks is now significantly lower, which makes it a lot easier to use Ansible's diff features, and also speeds up the playbook.

The PR is probably easiest to understand by looking at the individual commits, since they should all be completely self-contained.  One of the commits (the one to refactor common-server) got quite big, though.


playbooks/roles/backup-to-tarsnap/tasks/main.yml
```diff
@@ -32,9 +32,6 @@
     group: root
     mode: 0500
 
-- name: Check we can access tarsnap and run fsck
-  command: tarsnap --cachedir {{ TARSNAP_CACHE }} --keyfile {{ TARSNAP_KEY_REMOTE_LOCATION }} --fsck
```
UmanShahzad: Oh gosh thank you!

playbooks/roles/sanity-checker/tasks/main.yml
```diff
@@ -20,6 +20,3 @@
     job: "{{ SANITY_CHECK_SCRIPT_LOCATION }} {{ SANITY_CHECK_CONFIG }} > /dev/null 2>&1"
     minute: "10,20,30,40,50"
     name: "sanity checker for: {{ SANITY_CHECK_SUBJECT }}"
-
-- name: Run sanity check
-  command: "{{ SANITY_CHECK_SCRIPT_LOCATION }} {{ SANITY_CHECK_CONFIG }}"
```
UmanShahzad: Good idea here, it seemed out of place.

playbooks/group_vars/backup-pruner/public.yml
```diff
@@ -1,19 +1,2 @@
+COMMON_SERVER_NO_BACKUPS: true
 COMMON_SERVER_INSTALL_CONSUL: false
-
-TARSNAPPER_CONF: /etc/tarsnapper.yml
```
UmanShahzad: Was this completely unused?

smarnach: An `rgrep` in the secrets repo yielded only this single mesh, so yes, it was completely unused, including by third-party roles.

PR #29, opened 2018-05-03
Pin Filebeat's Elastic repo for Elasticsearch servers & tag everything

UmanShahzad:
When deploying Filebeat to all servers, I noticed I should probably pin the Elastic repo for Elasticsearch servers so they don't upgrade their Elasticsearch service accidentally to the latest version given by the Elastic 6.x apt repo.

Also, in order to run Filebeat alone on all servers without having to comment things out (which could cause issues with missing variables) and without having to use `--start-at-task` all the time (which requires us to eventually cancel the playbook mid-way), I tagged Filebeat with `'filebeat'` and ran `ansible-playbooks -vv deploy/playbooks/deploy-all.yml --tags "filebeat"` and only Filebeat ran on all servers without a problem. I then expanded this idea to every role in `common-server` plus all roles in playbooks.

### DIFF
--- a/playbooks/accounting.yml
+++ b/playbooks/accounting.yml
@@ -6,5 +6,8 @@
   hosts: accounting
   become: true
   roles:
-    - common-server
-    - accounting
+    - role: common-server
+      tags: 'common-server'
+
+    - role: accounting
+      tags: 'accounting'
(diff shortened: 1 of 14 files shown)


UmanShahzad:
@smarnach Yeah there needs to be a way to just say '--do-role' or something, but oh well.

PR #30, opened 2018-05-03
Backup Postgres every 3 hours rather than 1.

UmanShahzad:
Doing it every 1 hour makes it hard to deploy changes with migrations, since the backup takes so long (basically an hour or more).


UmanShahzad:
@smarnach We got the :+1: to go for 3h on Postgres -- can you double-check that this config will make sense? I have not deployed it yet.

playbooks/group_vars/postgres/public.yml
```diff
@@ -11,6 +11,9 @@ TARSNAP_BACKUP_POST_SCRIPT: "/usr/local/sbin/backup-post.sh"
 TARSNAP_BACKUP_FOLDERS: "{{ POSTGRES_SERVER_BACKUP_DIR }}"
 TARSNAP_ARCHIVE_NAME: "{{ POSTGRES_SERVER_DOMAIN }}"
 
+# Run every 3 hours.
+TARSNAP_BACKUP_HOUR: '0,3,6,9,12,15,18,21'
```
smarnach: How about using `*/3` instead?  We wouldn't control the exact times of the backup anymore, but I don't think we need to.

PR #34, opened 2018-05-07
Use common nginx roles for Ocim.

smarnach:
This makes use of the generic nginx and certbot roles for the nginx configuration of Ocim.  This reduces duplication and simplifies the code.  The new configuration isn't completely equivalent to the old one, e.g.

* The proxy buffer settings have been removed, and proxy buffering is completely disabled now.  There is no advantage to proxy buffering for us, since we don't use any nginx features that depend on it.
* The forwards from `www.` to the domain name without `www.` has been removed, since we don't have DNS records for these names anyway, so they were pointless.
* I removed the `proxy_connect_timeout` setting.  If all workers are so busy that they can't even accept the connection, then it's better to let the request fail, to reduce the load on the workers.

I also introduced a tag to make diffing the `.env` files easier, and updated the documentation accordingly.

This PR is already deployed to Ocim stage.


playbooks/roles/ocim/meta/main.yml
```diff
@@ -1,3 +1,20 @@
 dependencies:
+
   - name: pmbauer.tarsnap
     when: "{{ OPENCRAFT_BACKUP_SWIFT_ENABLED }}"
+
+  - name: nginx-proxy
```
UmanShahzad: Is `allow_duplicates` true by default? If not I believe we'll need to add it into `nginx-proxy`'s meta.

smarnach: As always, Ansible semantics are weird.  Roles with different parameters won't be deduplicated, regardless of `allow_duplicates`, and running nginx-proxy twice with the same parameters is indeed pointless, so we don't need that flag.  Furthermore, it works, and also worked for the prometheus role, which includes nginx-proxy twice as well.

PR #35, opened 2018-05-07
Copy nginx reload handler from common-server to nginx-proxy.

smarnach:
The handler is also needed by the appserver playbook, where the common-server role is not available.

### DIFF
--- a/playbooks/roles/nginx-proxy/handlers/main.yml
+++ b/playbooks/roles/nginx-proxy/handlers/main.yml
@@ -1,0 +1,5 @@
+---
+- name: reload nginx
+  service:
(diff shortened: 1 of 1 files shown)


smarnach:
@xitij2000 Could you please take a quick look?  This should fix the issue you reported on Mattermost.

smarnach:
@xitij2000 Merging now.  You can still review it after merging, but it will be a lot easier to test once merged, and it will probably save other people some trouble.

CC @UmanShahzad You were right with your comment after all. :)

xitij2000:
@smarnach I can understand the hurry here! This looks good to me.

PR #38, opened 2018-05-17
Install service-specific exporters.

smarnach:
Test instructions:

1. Go to https://prometheus.net.opencraft.hosting/targets. The default username is 'prometheus' and you should have the password.
2. Ensure that the HAProxy, MongoDB, and MySQL exporters show as being healthy targets (not red). (Note: you will see one of the HAProxy ones, haproxy-stage.opencraft.hosting, being red for reasons discussed on the ticket).
3. Go to https://haproxy-stage-3.net.opencraft.hosting:19101/metrics, https://mongodb-stage.opencraft.hosting:19001/metrics, and https://mysql-im-2-slave2.net.opencraft.hosting:19104/metrics. The default username is 'node-exporter' and you should have the password. Let your browser skip verifying the SSL certificate -- it's not provided by a recognized certificate authority.
4. For each of the above, ensure that you see a ton of variables somewhere that seem related to the particular service.
5. Go to https://prometheus.net.opencraft.hosting:3443/dashboards and login with credentials provided to you.
6. For each of MySQL, MongoDB, and HAProxy that you see on the list, open them up in new tabs. Take a look at each, and ensure the graphs are working (the data provided is literally the data seen from steps 3/4).

### DIFF
--- a/playbooks/roles/mongodb/meta/main.yml
+++ b/playbooks/roles/mongodb/meta/main.yml
@@ -1,0 +1,12 @@
+---
+dependencies:
+  - name: nginx-proxy
+    # The username is a bit misleading – should actually be "prometheus", since it's the server
+    # authenticating, and the same username is used for scraping other exporters, but it's not worth
+    # the hassle to change it.
+    NGINX_PROXY_USERNAME: node-exporter
+    NGINX_PROXY_PASSWORD: "{{ MONGODB_EXPORTER_PASSWORD }}"
+    NGINX_PROXY_AUTH_DOMAIN: Prometheus MongoDB Exporter
+    NGINX_PROXY_CONFIG_PATH: /etc/nginx/sites-available/mongodb-exporter
+    NGINX_PROXY_PUBLIC_PORT: 19001
+    NGINX_PROXY_INTERNAL_PORT: 9001
--- a/playbooks/roles/mysql/defaults/main.yml
+++ b/playbooks/roles/mysql/defaults/main.yml
@@ -1,9 +1,41 @@
 ---
 
-# MYSQL_OPENSTACK_DB_DEVICE --- volume location e.g.: ``/dev/vdb1``
-MYSQL_OPENSTACK_DB_DEVICE: "/dev/vdb1"
+# GENERAL #####################################################################
 
-MYSQL_SERVER_BACKUP_DIR: '/var/lib/mysql/backup'
-
+MYSQL_OPENSTACK_DB_DEVICE: /dev/vdb1
+MYSQL_SERVER_BACKUP_DIR: /var/lib/mysql/backup
 MYSQL_OPEN_FILES_LIMIT: 65536
 MYSQL_PROCESSES_LIMIT: 1024
+
+# PROMETHEUS ##################################################################
+
+MYSQL_EXPORTER_ENABLED: true
+MYSQL_EXPORTER_VERSION: 0.10.0
+MYSQL_EXPORTER_BASENAME: "mysqld_exporter-{{ MYSQL_EXPORTER_VERSION }}.linux-amd64"
+MYSQL_EXPORTER_BINARY: "/opt/{{ MYSQL_EXPORTER_BASENAME }}/mysqld_exporter"
+MYSQL_EXPORTER_ARCHIVE: "{{ MYSQL_EXPORTER_BASENAME }}.tar.gz"
+MYSQL_EXPORTER_DOWNLOAD_URL: "https://github.com/prometheus/mysqld_exporter/releases/download/v{{ MYSQL_EXPORTER_VERSION }}/{{ MYSQL_EXPORTER_ARCHIVE }}"
+MYSQL_EXPORTER_PASSWORD: null
+MYSQL_EXPORTER_MYSQL_PASSWORD: null
+
+# We use this by default because when you have a little too many databases,
+# the exporter takes ~40-50min to enter a listening state, and always times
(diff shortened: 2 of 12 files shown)


playbooks/roles/mysql/tasks/main.yml
```diff
@@ -1,17 +1,29 @@
+---
+
 # edX is currently using MySQL 5.6, which is not available on Xenial by default.
 - name: Add PPA for MySQL 5.6
   apt_repository:
     repo: "ppa:ondrej/mysql-5.6"
     update_cache: yes
```
smarnach: The MySQL role currently does not run, since it does not install the correct lib for Python 3.  This is a bug in the MySQL role we are using, and it is fixed in the latest version.  However, upgrading to the latest version requires more thorough testing, so I included this task here:
```
# This task can be removed once we move to the latest version of the geerlingguy.mysql role.  The
# version we are currently using only installs the Python2 version of python-mysqldb.
- name: Install python3-mysqldb package
  apt:
    name: python3-mysqldb
```
Note that it is not safe to run the MySQL role against the currently running servers, since setting up the replication is done in multiple steps, and running the role would run the earlier steps again – see Josh's comments on the ticket.  This means this fix is not particularly important, since we can't run the role anyway.  In the long run this needs to be sorted out, but that's out of scope for the current work.

playbooks/roles/mongodb/meta/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+dependencies:
+  - name: nginx-proxy
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/roles/mysql/meta/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+dependencies:
+  - name: nginx-proxy
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

playbooks/roles/mysql/tasks/main.yml
UmanShahzad: Ok, thanks for this info. I'll put that task there, and also be careful to not run the full role, but rather only step through it to install the nginx proxy and the exporter.

smarnach: Actually, to be more specific, I believe our `mysql` role is safe to run, but our fork of the external `geerlingguy.mysql` role isn't.  Please double-check this info before running against the production servers. :)

UmanShahzad: If you noticed that even one of them wasn't safe, that's enough for me :p I won't run the full roles, but rather only use `--step` to commit the tasks that that we want, e.g. the nginx-proxy and the exporter parts. I did that earlier for HAProxy stage when testing so it should be good.

playbooks/roles/mysql/defaults/main.yml
```diff
@@ -1,9 +1,41 @@
...
+
+# We use this by default because when you have a little too many databases,
+# the exporter takes ~40-50min to enter a listening state, and always times
+# out through the reverse proxy to actually retrieve these metrics.
+# It is uncertain whether this can actually be fixed in the exporter -- it
+# may just be a limitation of having too many databases in MySQL.
+MYSQL_EXPORTER_EXTRA_OPTS: "-collect.info_schema.tables=false"
+
+MYSQL_EXPORTER_MYSQL_USERS:
+  - name: exporter
```
symbolist: Nit: Used in task "Create MySQL exporter environment file" so could be a variable.

UmanShahzad: We've been trying to stay away from making something a variable unless we foresee it changing. In this case the password could definitely change but we don't see the username changing so that's why it isn't a variable.

playbooks/roles/mysql/meta/main.yml
```diff
@@ -0,0 +1,14 @@
...
+    when: MYSQL_EXPORTER_ENABLED
+    # The username is a bit misleading – should actually be "prometheus", since it's the server
+    # authenticating, and the same username is used for scraping other exporters, but it's not worth
+    # the hassle to change it.
+    NGINX_PROXY_USERNAME: node-exporter
+    NGINX_PROXY_PASSWORD: "{{ MYSQL_EXPORTER_PASSWORD }}"
+    NGINX_PROXY_AUTH_DOMAIN: Prometheus MySQL Exporter
+    NGINX_PROXY_CONFIG_PATH: /etc/nginx/sites-available/mysql-exporter
+    NGINX_PROXY_PUBLIC_PORT: 19104
+    NGINX_PROXY_INTERNAL_PORT: 9104
```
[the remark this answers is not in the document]
UmanShahzad: Similar comment as above, we actually don't see us changing this port for any reason. In fact changing it would also mean changing the public port to match it but with a prefixed "1", and it would also mean changing the security group on every server using this to open ingress on the proxy public port, which is just way too much work. So for that reason we simply don't bother making it a variable.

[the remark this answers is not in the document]
UmanShahzad: Note that it's the same case for MongoDB/HAProxy -- the ports are not variables, for all the same reasons, even though they are reused.

UmanShahzad:
@symbolist Thanks!

playbooks/roles/mysql/defaults/main.yml
symbolist: Ah okay! I was thinking of variables as performing the role of constants too. Thanks for the explanation.

PR #40, opened 2018-05-29
Add Mattermail role

viadanna:
This creates a new role to install Mattermail which is added to the chat.yml playbook.


playbooks/roles/mattermail/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
+# Connections settings for Mattermost
+MATTERMOST_URL: https://mattermost.example.com
+MATTERMOST_TEAM: opencraft
+MATTERMOST_LOGIN: mattermail
+MATTERMOST_PASSWORD: set-me-please
```
UmanShahzad: This variable isn't actually used, and to be honest a lot of these variable names are pretty ambiguous, e.g. what is the difference between `MATTERMOST_PASS` and `MATTERMOST_PASSWORD`?

Also, some variables, like `MATTERMOST_URL`, would I think basically be redefinitions of the existing `MATTERMOST_HOSTNAME` from the `mattermost` role. You already created a separate role at this point, but since Mattermail is so connected to Mattermost itself, do you think it'd make sense to merge this stuff into the existing `mattermost` role, with a flag that sets whether to install Mattermail or not? It can be brought in at the end of the `mattermost` role's task steps with `import_tasks` according to whether that conditional is set, if you decide to do this.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    home: "{{ MATTERMAIL_HOME }}"
+
+- name: Install git package
+  apt:
+    name: git
+    update_cache: yes
+
+- name: Install golang package
+  apt:
+    name: golang
```
UmanShahzad: You can merge this with the above as a list, e.g.:

```yml
- name: Install apt packages
  apt:
    name:
      - git
      - golang
    update_cache: yes
```

Although, I think we always install git through `COMMON_SERVER_DEPENDENCIES` and thus the `common-server` role, but up to you if you want to keep it.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: build Mattermail binary
+  command: go get {{ MATTERMAIL_REPO }}
```
UmanShahzad: Damn, there doesn't seem to be a module for this, which is surprising.

UmanShahzad: Also maybe you want to use `args.creates` (see [these examples](http://docs.ansible.com/ansible/latest/modules/command_module.html#examples)) so that this doesn't always run? Same for any other `command` modules used here.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    GOPATH: /opt/mattermail
+
+- name: create config
+  template:
+    src: config.json
+    dest: /opt/mattermail/config.json
+  become_user: mattermail
+
+- name: create data dir
+  command: mkdir {{ MATTERMAIL_HOME }}/data
```
UmanShahzad: You should use the [file module](http://docs.ansible.com/ansible/latest/modules/file_module.html) for making directories. It would basically be:

```yml
- name: Create Mattermail data directory
  file:
    path: "{{ MATTERMAIL_HOME }}/data"
    state: directory
    user: mattermail
    group: mattermail
```

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: Install golang package
+  apt:
+    name: golang
+
+- name: build Mattermail binary
+  command: go get {{ MATTERMAIL_REPO }}
+  become_user: mattermail
+  environment:
+    GOPATH: /opt/mattermail
```
UmanShahzad: Use `"{{ MATTERMAIL_HOME }}"`.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: create config
+  template:
+    src: config.json
+    dest: /opt/mattermail/config.json
```
UmanShahzad: Use `"{{ MATTERMAIL_HOME }}/config.json"`.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
+- name: create mattermail group
+  group:
+    name: mattermail
+
+- name: create mattermail user
+  user:
```
UmanShahzad: I think you can add `system: yes` here as well since this is really just a system account.

playbooks/roles/mattermail/tasks/main.yml
viadanna: Good tip!

playbooks/roles/mattermail/tasks/main.yml
viadanna: Whoops

playbooks/roles/mattermail/tasks/main.yml
viadanna: Yeah, I kept it for a test playbook on Vagrant. I'll test the chat playbook and remove this.

playbooks/roles/mattermail/defaults/main.yml
viadanna: I wanted to make this as independent as possible since Braden mentioned writing a Procmail filter to accomplish this in the future.

playbooks/roles/mattermail/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
...
+MATTERMOST_PASS: set-me-please
+MATTERMOST_CHANNEL: hosting-subteam
+
+# Mattermail settings to fetch emails
+MATTERMAIL_SERVER: imap.example.com:143
+MATTERMAIL_USERNAME: urgent
+MATTERMAIL_PASSWORD: set-me-please
+
+# Settings that can be left at their default values
+MATTERMAIL_REPO: github.com/rodcorsi/mattermail
```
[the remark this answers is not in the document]
xitij2000: Or if they have release tags set up you can set that up, and we can update the tag when there is a new release?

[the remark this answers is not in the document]
viadanna: Unfortunately the `go get` command used to fetch the repository and build the binary only supports fetching the `master`

playbooks/roles/mattermail/defaults/main.yml
UmanShahzad: I haven't looked into Procmail and likely won't have time, but wouldn't utilizing it mean mostly a complete replacement of most of this role anyway?

playbooks/roles/mattermail/templates/config.json
```diff
@@ -0,0 +1,22 @@
+{
+    "Directory": "./data/",
+    "Profiles":[
```
UmanShahzad: Nit: space between `:` and `[`.

playbooks/roles/mattermail/templates/config.json
```diff
@@ -0,0 +1,22 @@
...
+        {
+            "Name":     "{{ MATTERMOST_LOGIN }}",
```
UmanShahzad: Nit: I'm not sure if this a standard thing or not to do in JSON, but are the extra spaces really warranted?

playbooks/roles/mattermail/templates/config.json
```diff
@@ -0,0 +1,22 @@
...
+
+            "Mattermost":{
+                "Server":   "{{ MATTERMOST_URL }}",
+                "Team":     "{{ MATTERMOST_TEAM }}",
+                "User":     "{{ MATTERMOST_LOGIN }}",
+                "Password": "{{ MATTERMOST_PASSWORD }}"
+            }
+        }
+    ]
+}
```
UmanShahzad: Nit: need extra newline.

playbooks/roles/mattermail/templates/mattermail.service
```diff
@@ -0,0 +1,17 @@
+# mattermail
+[Unit]
+Description=mattermail server
+After=network-online.target
+
+[Service]
+Type=simple
+WorkingDirectory={{ MATTERMAIL_HOME }}
```
UmanShahzad: Nit: is this needed if we're already specifying `{{ MATTERMAIL_HOME }}` before `/bin/mattermail` right below?

playbooks/roles/mattermail/templates/config.json
```diff
@@ -0,0 +1,22 @@
+{
+    "Directory": "./data/",
+    "Profiles":[
+        {
+            "Name":     "{{ MATTERMOST_LOGIN }}",
+            "Channels": ["{{ MATTERMOST_CHANNEL }}"],
+
+            "Email":{
```
UmanShahzad: Nit: space between `:` and `{`.

playbooks/roles/mattermail/templates/config.json
```diff
@@ -0,0 +1,22 @@
...
+                "ImapServer": "{{ EMAIL_SERVER }}",
+                "Username":   "{{ EMAIL_USERNAME }}",
+                "Password":   "{{ EMAIL_PASSWORD }}"
+            },
+
+            "Mattermost":{
```
UmanShahzad: Nit: space between `:` and `{`.

playbooks/roles/mattermail/templates/config.json
UmanShahzad: I see that you got this from the README file on their repo. I guess they added spaces for easy initial demonstration.

playbooks/roles/mattermail/templates/mattermail.service
UmanShahzad: I see you got this from the README on their repo. Still, I wonder if this is necessary.

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+  command: go get {{ MATTERMAIL_REPO }}
+  args:
+    creates: "{{ MATTERMAIL_HOME }}/bin/mattermail"
+  become_user: mattermail
+  environment:
+    GOPATH: "{{ MATTERMAIL_HOME }}"
+
+- name: create config
+  template:
+    src: config.json
```
xitij2000: nit: use the .j2 extension for templates

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+- name: create data dir
+  file:
+    path: "{{ MATTERMAIL_HOME }}/data"
+    state: directory
+    owner: mattermail
+    group: mattermail
+
+- name: create systemd service
+  template:
+    src: mattermail.service
```
xitij2000: nit: same as above, use the .j2 extension

playbooks/roles/mattermail/defaults/main.yml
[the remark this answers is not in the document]
UmanShahzad: I looked at this again, and just remembered that since it's a Go package, it will 100% already upload the binary for easy download, if it's well maintained that is. And of course, it does: https://github.com/rodcorsi/mattermail/releases/download/3.5.0/mattermail-3.5.0.linux.am64.tar.gz

So we should add this as a variable URL, and insert the version as another variable as necessary. Then we can use `unarchive`, similar to [this example](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/consul/tasks/main.yml#L33-L40).

But then again, about security, I don't know enough about Go to know if these binaries are somehow sandboxed or something -- I'm guessing not? (Even then we can't be sure we're getting a Go binary). If it's just a matter of trust (e.g., we download the binary for Consul because we trust Hashicorp wouldn't give us something malicious), then I guess we can't assume trust for Mattermail, and would probably need our own repo...

@xitij2000 @viadanna what do you guys think? cc @bradenmacdonald

playbooks/roles/mattermail/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+  apt:
+    name:
+      - git
+      - golang
+    update_cache: yes
+
+- name: build Mattermail binary
+  command: go get {{ MATTERMAIL_REPO }}
+  args:
+    creates: "{{ MATTERMAIL_HOME }}/bin/mattermail"
```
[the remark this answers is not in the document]
UmanShahzad: Ah, that's right. But I still don't think we should be re-building the binary each time, that may cause issues for a running Mattermail, or maybe not? [This section from the Mattermost role](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/mattermost/tasks/mattermost.yml#L40-L106) may provide some help in what to do here. It's kind of hacky, and the other way is to use Ansible facts, but I'm not sure how much time we have to implement that.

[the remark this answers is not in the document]
xitij2000: One way is to store the git commit id in a file in MATTERMAIL_HOME and compare that to the current commit it to decide if an update it needed. It doesn't need to be a complex thing, you can use use ``set_fact`` I think.

playbooks/roles/mattermail/defaults/main.yml
viadanna: @xitij2000 That's confusing alright. For MATTERMOST_URL both should be the same, so is it safe to just reuse the values from another role?
@UmanShahzad That's right, which is why I thought of keeping a separate role instead of integrating with Mattermost role.

playbooks/roles/mattermail/templates/config.json
viadanna: That's how Go formats indented code to improve readability. Not very orthodox, so I'll remove extra spaces.

playbooks/roles/mattermail/templates/mattermail.service
viadanna: Yep, I agree. Probably unneeded, I'll check if any issues arises without it.

playbooks/roles/mattermail/tasks/main.yml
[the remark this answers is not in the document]
viadanna: I didn't expect this to be used for upgrading as well. There shouldn't be any issue replacing the binary while Mattermail is running, but the service should be restarted instead of enabled and started.
I'll evaluate storing the git hash anyway.

playbooks/roles/mattermail/tasks/main.yml
viadanna: 👍

playbooks/roles/mattermail/tasks/main.yml
viadanna: 👍

playbooks/roles/mattermail/defaults/main.yml
xitij2000: I just mean to do that for variable names that convey the same information but need two differnt names for some reason.

playbooks/roles/mattermail/defaults/main.yml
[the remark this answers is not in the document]
xitij2000: @UmanShahzad  Code isn't any safer than a binary if it hasn't been audited :-) Do we fully 100% trust the code to begin with? Because the code and binary are from the same source.

playbooks/roles/mattermail/tasks/main.yml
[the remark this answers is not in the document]
xitij2000: Enabling is idempotent if it's already enabled nothing will happen. I don't think we will be deploying a lot of new Matermost servers, we'll probably be running the playbook on the same server again and again. Ideally, the upgrades would be controlled using some variable with the version number, like @UmanShahzad  suggested above with the binary url.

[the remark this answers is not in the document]
UmanShahzad: You can add a `notify: restart mattermail` and a handler to restart it on this step or whatever seems right, so the systemd service will get restarted.

- [Example notify](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/consul/tasks/main.yml#L48)
- [Example handler](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/consul/handlers/main.yml)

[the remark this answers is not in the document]
viadanna: @xitij2000 The idea of having our own repository for Mattermail would be that we control the master branch (which would be used to build the binary using `go get`)  thus removing any responsibility from the playbook to handle versions, simplifying configuration. What do you think?

@UmanShahzad It seems that having it always enable and restart at the end also works and seems simpler:
```yaml
- name: start and enable the mattermail service
  service:
    name: mattermail
    state: restarted
    enabled: yes
```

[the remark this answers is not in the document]
xitij2000: Sure, having our own fork is OK to if we don't go for the binary approach.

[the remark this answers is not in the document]
xitij2000: About having a notifier, that makes sense but only if we're only running the update if there is a version upgrade. Otherwise it will always run the handler so might as well just have a regular task.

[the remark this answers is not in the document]
UmanShahzad: @viadanna The notifier would only get run if the task got run successfully, and in general it's not good practice to restart the service if it isn't needed; imagine doing this for MySQL, for example, it would restart the service and bring down all connections anytime we run the playbook, which isn't ideal. I agree it isn't that critical here, but we should still be consistent / follow best practices.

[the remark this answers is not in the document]
UmanShahzad: @xitij2000 Yep, I would only say add the notifier if we plan on implementing the ability to skip the install step if it isn't needed. Otherwise it would indeed always get run and would make no difference.

viadanna:
@xitij2000 I opted for using our own fork and checking the hash before updating/restarting. What do you think?

xitij2000:
@viadanna  Looks good! I'm not very familiar with ``go`` so perhaps what I'm suggesting is infeasible or inadvisable, but is the following approach possible:

Clone/update the repository to a local folder
If the update resulted in a change, use ``go install`` instead of ``go get`` to install the package.

I think this way we don't need to track the hash since that's automatical with git. 

Additionally, even if we use ``go get`` wouldn't you need to use ``go get -u`` to update?

Once the above is addressed I think it's good to merge. I doubt we have any staging mattermost server, so I guess I'll have to test this in a local VM. Won't find time for that today sadly.

viadanna:
> Clone/update the repository to a local folder
> If the update resulted in a change, use go install instead of go get to install the package.
> 
> I think this way we don't need to track the hash since that's automatical with git.

That's perfectly feasible, I just went with a simpler solution as Go is pretty strict with relation to path names and the GOPATH variable. Using `go get` means less variables and conditionals being implemented by the role.

> Additionally, even if we use go get wouldn't you need to use go get -u to update?

You are correct sir, completely missed that.

> Once the above is addressed I think it's good to merge. I doubt we have any staging mattermost server, so I guess I'll have to test this in a local VM. Won't find time for that today sadly.

I've been using a bare Vagrant box with `config.vm.box = "ubuntu/xenial64"` and it's solid for testing this. The snapshotting feature helps a lot.

bradenmacdonald:
I can't see the thread for the discussion about installing from binaries, but that's a big advantage of go. - installation is as simple as copying a single file into /usr/loca/bin. So if it's an official binary and we verify the hash before installing it, I think that's great.

See e.g. https://www.terraform.io/intro/getting-started/install.html

viadanna:
Hey @bradenmacdonald 

The thread is [here](https://github.com/open-craft/ansible-playbooks/pull/40/files/430b38e903ee7f984ccedb9393c13377d5614722#diff-4fc14d24e9524c9ae7c68cd821071d46).

I opted to using Git and Go to build from source locally as it compiles pretty much instantaneously and the tooling has a small footprint.

Do you recommend changing the the implementation to use the binaries provided in Github?

UmanShahzad:
The only thing I have against it is, I don't think the inventor of Mattermail is some big official group that we can just 'trust' necessarily. So for the same reason we used to not use the `master` branch for the external ansible roles and rather use commits in case there is malicious intent, I'd say we should use our own branch and just compile.

playbooks/roles/mattermail/templates/config.json.j2
```diff
@@ -0,0 +1,22 @@
+{
+    "Directory": "{{ MATTERMAIL_HOME }}/data/",
+    "Profiles": [
```
UmanShahzad: So we don't seem to have a way to actually filter the things we want, because the `Filter` key is not present. See https://github.com/rodcorsi/mattermail#filter.

Additionally, I think we need to use the `UseAPIv3` key here from https://github.com/rodcorsi/mattermail#mattermost.

Please check that README to see if anything else we may need has been missed.

viadanna: Yeah, I expected a mail account getting only urgent emails, but filtering is available if that's not the case. There's also a redirect by subject option but there's little documentation on that.

I agree that UseAPIV3 might be necessary, let me know if you want me to make the change @xitij2000

xitij2000: @UmanShahzad  @viadanna  I don't see why ``UseAPIV3`` is necessary at all. We are already using Mattermost 4.x. 

Am I missing something here?

UmanShahzad:
@viadanna Just a small request for when we deploy: can you rebase this on master to make sure we have all other variables fully up-to-date? Or @xitij2000 since you will be deploying maybe you can do that, whatever works.

xitij2000:
@viadanna  Could you squash the comments and merge?

viadanna:
Sure @xitij2000

PR #41, opened 2018-05-29
Increment tarsnap version to 1.0.39

kaizoku:
This updates Tarsnap to the newest version (1.0.39) as of May 29th, 2018.
I've tested and deployed this on our production and staging servers, and we should use this on new servers going forward.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/kamaln7/ansible-swapfile
 [submodule "playbooks/roles/pmbauer.tarsnap"]
 	path = playbooks/roles/pmbauer.tarsnap
-	url = https://github.com/pmbauer/ansible-tarsnap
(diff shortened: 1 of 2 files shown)


UmanShahzad:
@kaizoku This is changing `group_vars/all/public.yml`, which doesn't exist anymore. Are you sure you have the latest ansible-playbooks version? See https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/group_vars/all/public.yml#L9.

I worry that if you ran the playbook with the older version, if something weird happened :s I think nothing did, but yeah.

kaizoku:
Ah whoops, I didn't realize that repo was behind. I rebased against master, and it applied the changes to the moved file, `playbooks/group_vars/all/public.yml`.
(And when I ran the playbook, I had made the same change in the `deploy/` submodule, so it wasn't this old version)

Also I added a commit which switches to using our fork of pmbauer.tarsnap.

PR #43, opened 2018-06-06
Add logrotation configuration, set mongodb_daemon variable to fix logrotation on MongoDB servers

rocioar:
The original mongodb playbook is using mongodb_package to define what daemon is running Mongo, and with that is guessing the log name. See: https://github.com/UnderGreen/ansible-role-mongodb/blob/82a303001f411ac6b835c348a9de7e6319a4864e/defaults/main.yml#L3

This PR sets mongodb_package to MongoDB (by default it was mongodb-org, which is the version that's installed when you download it directly from Mongo instead of apt)


rocioar:
@smarnach this should fix both issues, the one with stage (there was an issue with the name of the logfile being mongod.log instead of mongodb.log, that was caused by the daemon name variable), and also the issues with the replicas, they didn't have a logrotate configuration installed at all.

rocioar:
I've tested this and it seems to be doing the proper changes, the only thing now, is the issue with pip3 and the replicas

playbooks/roles/mongodb/files/logrotate
```diff
@@ -0,0 +1,8 @@
+/var/lib/mongodb/log/mongodb.log {
```
UmanShahzad: Nit: this should be configurable and then become a Jinja2 template

rocioar: I was following the way we do it for prometheus. Will fix.

smarnach: @rocioar For Prometheus, I didn't make the log file location configurable at all.  In this case however, the location is already configurable, so it makes sense to use a template here.

rocioar:
@UmanShahzad @smarnach fixed :)

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -30,6 +30,15 @@
   with_items:
     - 27017
 
+- name: configure log rotation for the MongoDB logs
+  copy:
+    template:
```
UmanShahzad: I think there was a typo here. `copy` shouldn't be here.

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -30,6 +30,15 @@
...
+      src: logrotate.conf.j2
+      dest: /etc/logrotate.d/mongodb.conf
+      owner: root
+      group: root
```
UmanShahzad: `root` is the default `owner` and `group`, so these can be erased to keep it succinct.

rocioar:
@UmanShahzad the PR is ready for review again, I've addressed your feedback and fixed the issue with pip3 by using python2 as the interpreter instead.

playbooks/group_vars/all/public.yml
```diff
@@ -2,7 +2,7 @@
 
 # ANSIBLE #####################################################################
 
-ansible_python_interpreter: /usr/bin/python3
+ansible_python_interpreter: /usr/bin/python
```
UmanShahzad: I think we want to keep using python3 by default, but just use python2 for the MongoDB servers.

UmanShahzad:
@rocioar Thanks! One last issue, but I think it's good after that. Just to be sure, the playbook worked, right?

rocioar:
@UmanShahzad  fixed it, I've run it with the --check command, and it seems to do the right thing. Shall I proceed to running it for real now?

rocioar:
@UmanShahzad done, and worked fine 👍

UmanShahzad:
@rocioar Thanks! I will review soon then.

rocioar:
@UmanShahzad  run for both the secondaries, went well, merging the PR now.

UmanShahzad:
@rocioar I think Filebeat is doing funny things now, because it's picking up the gzipped file contents too, and logz.io is filled with garbage logs now.

The filebeat config is in the MongoDB group vars where you forced Python 2 usage, as this:

```yml
 filebeat_prospectors:
   - fields:
       type: "mongodb"
     paths:
       - "{{ mongodb_systemlog_path | dirname }}/*.log*"
```

We need to no longer capture `*.log*`, but rather only the actual log file. So maybe just `*.log` would do, but do confirm.

PR #44, opened 2018-06-06
Fix failing pre-backup script for LPD.

tomaszgy:
This PR fixes failing `lpd-pre-backup.sh` script.
Changes have been already applied to LPD stage and production.

cf. [OC-4365](https://tasks.opencraft.com/browse/OC-4365)

Signed-off-by: Tomasz Gargas <tomasz@opencraft.com>

### DIFF
--- a/playbooks/learner-profile-dashboard.yml
+++ b/playbooks/learner-profile-dashboard.yml
@@ -19,6 +19,10 @@
       # Backup of logs is initiated by logrotate, which incidentally
       # also creates data to be backed up.
       TARSNAP_CRONTAB_STATE: "absent"
+
+      # The backup lock is causing problems for LPD, since the lock is inherited by
+      # gunicorn workers, which don't release it unless they are stopped.  Since the lock
+      # is not needed for a daily backup anyway, we simply disable it.
       TARSNAP_BACKUP_GLOBAL_LOCK: false
 
     - role: learner-profile-dashboard
--- a/playbooks/learner-profile-dashboard.yml
+++ b/playbooks/learner-profile-dashboard.yml
@@ -19,6 +19,7 @@
       # Backup of logs is initiated by logrotate, which incidentally
       # also creates data to be backed up.
       TARSNAP_CRONTAB_STATE: "absent"
+      TARSNAP_BACKUP_GLOBAL_LOCK: false
 
     - role: learner-profile-dashboard
       tags: 'learner-profile-dashboard'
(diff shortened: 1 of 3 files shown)


playbooks/roles/learner-profile-dashboard/templates/lpd-pre-backup.sh
```diff
@@ -6,7 +6,9 @@ service gunicorn restart
 
 timestamp=$(date +%Y%m%dT%H:%M:%S)
 
-mv /var/log/lpd/debug.log-* "{{ LPD_LOG_DOWNLOAD_LOG_DIR }}"
+if [ -f /var/log/lpd/debug.log-* ]; then
+   mv /var/log/lpd/debug.log-* "{{ LPD_LOG_DOWNLOAD_LOG_DIR }}"
```
tomaszgy: @itsjeyd Here leaving `mv` without the `if` clause makes `mv` fail (and consequently, the whole script) when `/var/log/lpd/debug.log-*` doesn't match any file.

playbooks/learner-profile-dashboard.yml
```diff
@@ -19,6 +19,7 @@
       # Backup of logs is initiated by logrotate, which incidentally
       # also creates data to be backed up.
       TARSNAP_CRONTAB_STATE: "absent"
+      TARSNAP_BACKUP_GLOBAL_LOCK: false
```
itsjeyd: @tomaszgy Could you add the following comment above this line:

```
      # The backup lock is causing problems for LPD, since the lock is inherited by
      # gunicorn workers, which don't release it unless they are stopped.  Since the lock
      # is not needed for a daily backup anyway, we simply disable it.
```

The same info is already available in the [playbook](https://github.com/open-craft/ansible-playbooks/blob/35e9bdfe0fdb7e125806c356bd1d58a747118c49/playbooks/dalite.yml#L22-L24) for Dalite, but further down the line it probably won't be obvious where to look for it. So it would be worth repeating here :)

tomaszgy: Sure, done :)

tomaszgy:
@itsjeyd Thanks for the first round of review! This is ready for another round now :)

itsjeyd:
@tomaszgy I'll go ahead and merge this PR now; need to base deployment updates for [OC-4614](https://tasks.opencraft.com/browse/OC-4614) on the changes that it introduces.

itsjeyd:
@tomaszgy I think we'll need to follow up with a submodule hash update for ansible-secrets? It currently points to https://github.com/open-craft/ansible-playbooks/tree/35e9bdfe0fdb7e125806c356bd1d58a747118c49 but the most recent `master` is https://github.com/open-craft/ansible-playbooks/tree/5a8ae4a5a2d30d80940705f0b53c13d0137a968a.

PR #45, opened 2018-06-06
Add initial Nomad role

UmanShahzad:
This PR adds the relevant Nomad role, playbook, and common public variables needed to get a Nomad cluster running, including both the server and client nodes.

Test Instructions:

- Check certificates on any single Nomad server and client with `openssl x509 -noout -text -in cert.pem` -- ensure that the SAN list makes sense.
- On both a Nomad server and client, do `nomad server members` and ensure all Nomad servers for the cluster are there. Do `nomad node status` to see that all clients are there.
- Run the Nomad playbook afresh, without selecting any group -- see that it succeeds for all involved servers without issue.


playbooks/nomad.yml
```diff
@@ -0,0 +1,25 @@
...
+
+- name: Set up Nomad server
+  hosts: nomad-server
+  become: true
+  roles:
+    # Note: we don't run the common-server role because our Nomad servers
+    # are on the load balancing servers, so running it would cause issues
+    # like the hostname changing every now and then.
+    # We can technically skip such tasks with variables, but we don't need
+    # anything specific out of common-server that is tailored for Nomad.
```
smarnach: Maybe we should just run the common-server role for `hosts: all`?  However, this would make it more difficult to use the individual playbooks, so it's probably not a good idea.

playbooks/nomad.yml
```diff
@@ -0,0 +1,25 @@
...
+    - role: nomad
+      nomad_server: true
+      tags: ['nomad', 'nomad-server']
+
+- name: Set up Nomad client
+  hosts: nomad-client
+  become: true
+  roles:
+    - role: common
```
smarnach: What's this role?  I don't see it in this PR.

playbooks/roles/nomad/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
```
smarnach: In it's current form, this file isn't particularly useful. :)

playbooks/roles/nomad/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+nomad_client: false
+nomad_enable_tls: true
+nomad_enable_telemetry: true
+
+nomad_bin_dir: /usr/local/bin
+nomad_data_dir: /var/lib/nomad
+nomad_config_dir: /etc/nomad
+nomad_tls_dir: "{{ nomad_config_dir }}/tls"
+
+nomad_client_reserved_cpu: 500 # MHz
```
smarnach: This seems quite a bit – is this the recommened amount, or did you just make this up?  (Most schedulers I know use fractions of CPUs as a unit, which I find easier to handle, but I can see why Nomad chose to use MHz instead.)

playbooks/roles/nomad/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+
+dependencies: []
```
smarnach: We don't need this file if it's empty anyway.  It was only required for Ansible Galaxy.

playbooks/roles/nomad/README.md
UmanShahzad: Yeah I used `ansible-galaxy init` to generate this. This is one of the things to polish.

playbooks/roles/nomad/defaults/main.yml
UmanShahzad: I took this value straight from the document examples for lack of a better value, but it definitely did seem like too much to me too, and I planned on changing it. How much do you think I should put here though, that'll be required by things like Filebeat, Consul, the kernel, etc?

playbooks/roles/nomad/meta/main.yml
UmanShahzad: Yep, will delete when I start polishing the PR(s).

playbooks/nomad.yml
UmanShahzad: Woops, should be `common-server`.

playbooks/nomad.yml
UmanShahzad: Yeah I think that may cause issues we can't foresee right now, so I'll leave it as is for now.

playbooks/roles/nomad/defaults/main.yml
UmanShahzad: Thinking about it, I think we should just not reserve CPU at all, but reserve only memory instead. We can probably make do with other services, at least Filebeat & the Consul client agent, running a bit slower because of lack of CPU cycles being allocated to them, since I think only memory shortage would cause visible issues.

smarnach: Sounds good to me – I also expect memory to be the limiting resource, not CPU.  We can always change it after we gain some experience.

playbooks/group_vars/load-balancer/public.yml
```diff
@@ -17,8 +17,8 @@ SANITY_CHECK_LIVE_PORTS:
     message: "Load balancer does not respond on port 443."
 
 # TARSNAP #####################################################################
-
-TARSNAP_BACKUP_FOLDERS: "/etc/letsencrypt /etc/haproxy {{ consul_data_dir }}"
+  
+TARSNAP_BACKUP_FOLDERS: "/etc/letsencrypt /etc/haproxy {{ consul_data_dir }} /var/lib/nomad"
```
UmanShahzad: Note that this is a global variable for all load balancers, including ones not running Nomad masters like the current HAProxy prod/stage. So to prevent any errors without having to make a whole new group vars file and change up the SSH config a lot, I just went in and `mkdir /var/lib/nomad` on those servers.

.gitmodules
```diff
@@ -10,6 +10,7 @@
 [submodule "playbooks/roles/pmbauer.tarsnap"]
 	path = playbooks/roles/pmbauer.tarsnap
 	url = https://github.com/open-craft/ansible-tarsnap
+	branch = master
```
UmanShahzad: We made a fork recently to make a change needed to upgrade the tarsnap version. The upstream repo's last commit is December '16 so I'm not sure if it's worth trying to upstream that change.

playbooks/group_vars/nomad-client/public.yml
```diff
@@ -0,0 +1,16 @@
+---
+
+# COMMON ######################################################################
+
+COMMON_SERVER_NO_BACKUPS: true
+COMMON_SERVER_INSTALL_NODE_EXPORTER: false
```
UmanShahzad: See my `TODO` in the Nomad configuration file below.

playbooks/group_vars/nomad-client/public.yml
```diff
@@ -0,0 +1,16 @@
...
+COMMON_SERVER_INSTALL_SANITY_CHECKER: false
```
UmanShahzad: Because we're now phasing out sanity checker usage.

.gitmodules
smarnach: Many external roles simply do not seem worth it for us, and I guess this is one of them.  External roles are useful to quickly get started, and sometimes to do significant amounts of work for us, but once the overhead of configuring and adapting them outweighs their benefits we should move to our own roles, possibly based on the external role.  Installing tarsnap is really easy, and I think we should get rid of this external role in the long run.

playbooks/group_vars/load-balancer/public.yml
```diff
@@ -17,8 +17,8 @@ SANITY_CHECK_LIVE_PORTS:
     message: "Load balancer does not respond on port 443."
 
 # TARSNAP #####################################################################
-
-TARSNAP_BACKUP_FOLDERS: "/etc/letsencrypt /etc/haproxy {{ consul_data_dir }}"
+
```
smarnach: Trailing whitespace, please remove.

playbooks/group_vars/load-balancer/public.yml
smarnach: You could conditionally add the directory to the list, based on whether the host is in the "nomad" group, e.g.

    TARSNAP_BACKUP_FOLDERS: "/etc/letsencrypt /etc/haproxy {{ consul_data_dir }} {{ '/var/lib/nomad' if 'nomad' in group_names else '' }}"

(I did not test this, but something along this lines should work.)

playbooks/roles/nomad/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+
+- name: Download and unarchive the Nomad binary
+  unarchive:
+    src: "{{ nomad_url }}"
+    dest: "{{ nomad_bin_dir }}"
+    owner: nomad
+    group: nomad
+    mode: +x
+    remote_src: yes
+    creates: "{{ nomad_bin_dir }}/nomad"
```
smarnach: So this won't run again if we update the nomad version?  Might be worth adding a comment to the version number, so the next person trying to upgrade it doesn't get confused.

playbooks/group_vars/load-balancer/public.yml
UmanShahzad: Yep cool, that works!

playbooks/roles/nomad/tasks/main.yml
UmanShahzad: Will add a comment :+1:

PR #49, opened 2018-06-14
Fix filebeat log path for mongodb

rocioar:
More details: https://github.com/open-craft/ansible-playbooks/pull/43#issuecomment-397143857


rocioar:
I've applied it to the secondaries and stage servers, what about the master db? Shall I proceed too?

PR #53, opened 2018-07-02
Make a mattermail template without message and with @here.

UmanShahzad:
With this, we should get the following result:

```
"MailTemplate": "@here :incoming_envelope: _From: **{{.From}}**_\n>_{{.Subject}}_",
```

I have confirmed that with running the playbook with `--diff`.

### DIFF
--- a/playbooks/roles/mattermail/defaults/main.yml
+++ b/playbooks/roles/mattermail/defaults/main.yml
@@ -14,3 +14,6 @@ EMAIL_PASSWORD: set-me-please
 MATTERMAIL_REPO: github.com/open-craft/mattermail
 MATTERMAIL_BRANCH: master
 MATTERMAIL_HOME: /opt/mattermail
+
+# Misc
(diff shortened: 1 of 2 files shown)


playbooks/roles/mattermail/defaults/main.yml
```diff
@@ -14,3 +14,6 @@ EMAIL_PASSWORD: set-me-please
 MATTERMAIL_REPO: github.com/open-craft/mattermail
 MATTERMAIL_BRANCH: master
 MATTERMAIL_HOME: /opt/mattermail
+
+# Misc
+MATTERMAIL_MAIL_TEMPLATE: "@here :incoming_envelope: _From: **{{'{{'}}.From{{'}}'}}**_{{'\n'}}>_{{'{{'}}.Subject{{'}}'}}_"
```
viadanna: That's some crazy amount of braces 😄 

Good idea using a variable for helping with further changes.

UmanShahzad: Yeah there are some special characters in there that ansible looks for which were causing issues, so I found we could do it this way. There may be a better way but this works fine.

PR #56, opened 2018-07-11
Fix version hash of pmbauer.tarsnap role.

smarnach:
The old version hash does not exist in the repository anymore.  I replaced it by the hash of the latest master version instead.  This way, recursively checking out this repository works again.


smarnach:
@UmanShahzad I'm merging this PR immediately, since it's rather trivial, and I want to correct some hashes in ansible-secrets as well.

UmanShahzad:
@smarnach This seems to have broken Ocim deployments:

```
AppServer deploy failed: unhandled exception
  File "/var/www/opencraft/instance/models/openedx_appserver.py", line 490, in provision
    log, exit_code = self.run_ansible_playbooks()
  File "/var/www/opencraft/instance/models/mixins/ansible.py", line 124, in run_ansible_playbooks
    with open_repository(playbook.source_repo, ref=playbook.version) as configuration_repo:
  File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/var/www/opencraft/instance/repo.py", line 53, in open_repository
    repo.submodule_update()
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/repo/base.py", line 305, in submodule_update
    return RootModule(self).update(*args, **kwargs)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/objects/submodule/root.py", line 327, in update
    progress=progress, dry_run=dry_run, force=force_reset, keep_going=keep_going)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/objects/submodule/base.py", line 619, in update
    mrepo.head.reset(hexsha, index=True, working_tree=True)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/refs/head.py", line 76, in reset
    self.repo.git.reset(mode, commit, '--', paths, **kwargs)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/cmd.py", line 423, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/cmd.py", line 868, in _call_process
    return self.execute(call, **_kwargs)
  File "/var/www/.virtualenvs/opencraft/lib/python3.5/site-packages/git/cmd.py", line 686, in execute
    raise GitCommandError(command, status, stderr_value, stdout_value)
  cmdline: git reset --hard 7a625912409a0be2a3bf48967c60c5328ae2f3b2 --
  stderr: 'fatal: Could not parse object '7a625912409a0be2a3bf48967c60c5328ae2f3b2'.'
```

We had this issue before in fact which is why it is the way it is right now. I don't know yet how to achieve what you want and how to avoid this issue at the same time right now (just haven't given it much thought), but I'll undo this for now so the Ocim mass redeployment doesn't fail.

UmanShahzad:
@smarnach I made https://github.com/open-craft/ansible-playbooks/pull/57 to make this a role in the repo itself now.

smarnach:
@UmanShahzad Thanks for fixing this.  I didn't think about that this is also used by Ocim instances now, otherwise I wouldn't have merged without further testing.

I'm completely puzzled, though.  The submodule hash that was used before doesn't even exist in the repo.  How can this work?  And why did  it break when I replaced it with a hash that actually exists?

UmanShahzad:
@smarnach The latest `master` commit on open-craft/ansible-tarsnap shows to me as https://github.com/open-craft/ansible-tarsnap/commit/46032be1cbbecfbe1c77a6da24a891878216635e. I think we did something weird with the submodule and everybody's old commit became 7a625912409a0be2a3bf48967c60c5328ae2f3b2 which disappeared from the main line somehow. I was able to fix the hash with:

```bash
$ git submodule sync --recursive
$ git submodule update --init --recursive
```

when inside of ansible-playbooks. This fixes it on all of our director instances who have this repo checked out too.

smarnach:
@UmanShahzad My version of pmbauer.tarsnap still used the upstream repo rather than the fork, and weirdly continued using it even after completely removing the whole directory and updating submodules again.  The two repos have diverged, so the latest master commit in our repo does not exist in upstream and vice versa, which is why we saw different commits as non-existent.

UmanShahzad:
@smarnach Ah yes, that should explain it. Git submodules are weird...

PR #57, opened 2018-07-12
Move ansible-tarsnap into this repository.

UmanShahzad:
This should help us avoid the issues we're seeing with any issues with submodules for this role.


UmanShahzad:
@smarnach Will you be able to review this or should I just merge it? Ocim deploys are currently all failing, and the mass redeployment has been stalled, so maybe we should just merge it since it'd probably be better than what's currently going on.

playbooks/roles/tarsnap/tasks/tarsnap.yml
```diff
@@ -0,0 +1,70 @@
...
+  when: tarnsap_installed|failed
+  command: tar xzf /root/tarsnap-autoconf-{{ tarsnap_version }}.tgz chdir=/root creates=/root/tarsnap-autoconf-{{ tarsnap_version }}/COPYING
+
+- name: Configure Tarsnap for local build
+  when: tarnsap_installed|failed
+  command: ./configure chdir=/root/tarsnap-autoconf-{{ tarsnap_version }} creates=/root/tarsnap-autoconf-{{ tarsnap_version }}/Makefile
+
+- name: Build and install Tarsnap
+  when: tarnsap_installed|failed
+  command: make all install clean chdir=/root/tarsnap-autoconf-{{ tarsnap_version }}
```
smarnach: You seem to have removed the `creates` setting here, so this will now always be run, even when the tarsnap binary already exists.  Is there any specific reason why you did this?

kaizoku: Right @UmanShahzad, there's already a guard against running this excessively. If the correct version is already installed, `tarsnap_installed` will be set, and this won't run.

With the `creates` setting there was no way to update tarsnap, except I guess removing the binary before running the playbook, but that didn't seem like a good solution.

PR #61, opened 2018-07-17
Replace Crafty by the Mattermost autolink plugin.

smarnach:
The latest version of Mattermost does not support APIv3 anymore, which Crafty was still using.  Instead of porting to the new API version, it was quicker and nicer to port to the new autolink plugin.

This code mostly automates the installation and configuration of the plugin, though a few manual steps are still required.  The main problem is that the Mattermost configuration file is owned by Mattermost, not by Ansible, so Ansible can only apply careful changes.  We would like Ansible to own the autolink plugin configuration, which is part of the configuration file.  This would be possible with Ansible's `blockinfile` module.  However, that module requires marker comments to delimit the managed block, and Mattermost's JSON config file does not support comments.

It would be possible to implement something manually to fully automate this, but in the limited time I had available this was all I could do.


playbooks/roles/mattermost/templates/autolink.json.j2
```diff
@@ -0,0 +1,18 @@
+# The lines below should be included in the section PluginSettings -> Plugins -> mattermost-autolink
+# in the file config.json.  Since config.json is owned by Mattermost, and JSON does not support any
+# comments, it's difficult to automate this with Ansible.
```
UmanShahzad: But now that it's rendered and manually added to `config.json` already, couldn't we add it to the template anyway, to make it truly automated for future deployments?

I guess the main use of leaving it as a separate config file that has to be manually merged is if we foresee that `MATTERMOST_JIRA_DOMAINS` will change relatively often, forcing us to do something awkward to see the rendered result and change `config.json` manually anyway. If we foresee this variable changing then nevermind this comment.

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -32,15 +32,20 @@
...
         "SessionCacheInMinutes": 10,
+        "SessionIdleTimeoutInMinutes": 0,
         "WebsocketSecurePort": 443,
         "WebsocketPort": 80,
         "WebserverMode": "gzip",
         "EnableCustomEmoji": false,
         "EnableEmojiPicker": true,
+        "EnableGifPicker": false,
+        "GfycatApiKey": "2_KtH_W5",
+        "GfycatApiSecret": "3wLVZPiswc3DnaiaFoLkDvB4X0IV6CpMkj4tf2inJRsBY6-FnkT08zGmppWFgeof",
```
UmanShahzad: Is this supposed to be public?

playbooks/roles/mattermost/templates/autolink.json.j2
smarnach: @UmanShahzad We add a new project to our own Jira instance every few months, and this requires an update to `MATTERMOST_JIRA_DOMAINS` everytime.  I chose the design you reviewed to make these updates relatively painless – we just need to copy the few lines from the one file to the other.

However, based on your comment, I slightly modified the solution so it builds the full configuration on fresh installs – you are right that we should do that given that it's so simple now.  I had to turn the fake comment into a Jinja2 comment to prevent it from showing up in the output, since JSON files can't have comments.

playbooks/roles/mattermost/templates/config.json.j2
smarnach: No, it decidedly isn't.  No idea what I need to do now…

playbooks/roles/mattermost/templates/autolink.json.j2
smarnach: I tested the new version by moving `config.json` on the server out of the way, running the playbook to generate a new version and verifying that both files are essentially identical, which they are.

playbooks/roles/mattermost/templates/config.json.j2
smarnach: Well, it turns out this actually _is_ supposed to be public – these are [upstream's default values](https://github.com/mattermost/mattermost-server/blob/f2c180390599e66fee2f1a8c1a4ab52eea920c51/config/default.json#L48-L49).  I've got no idea what this is, but now I don't care anymore.

smarnach:
@UmanShahzad Thanks for the review!  I''ve addressed your comments, so I'm merging this now.

PR #63, opened 2018-07-18
Add a certbot hook to restart nginx on deploying renewed certificates

lgp171188:
This PR adds a certbot hook script `restart-nginx.sh` (which reloads the nginx service) to the `/etc/letsencrypt/renewal-hooks/deploy` directory. Verified that nginx uses the renewed certificate by doing a forced renewal with the `--force-renewal` flag and checking the validity of the certificate presented by nginx on connecting to port 19100.

### DIFF
--- a/playbooks/roles/certbot/tasks/main.yml
+++ b/playbooks/roles/certbot/tasks/main.yml
@@ -43,6 +43,17 @@
     path: /var/www/certbot
     state: directory
 
+- name: Create Certbot deploy hooks directory
+  file:
+    path: /etc/letsencrypt/renewal-hooks/deploy
+    state: directory
+
+- name: Copy the Certbot deploy hook to restart nginx on renewal
+  copy:
+    src: restart-nginx.sh
+    dest: /etc/letsencrypt/renewal-hooks/deploy/restart-nginx.sh
+    mode: 0755
+
 - name: generate SSL certificate with certbot
   shell: >-
     certbot certonly
--- a/playbooks/roles/certbot/files/restart-nginx.sh
+++ b/playbooks/roles/certbot/files/restart-nginx.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+systemctl reload nginx.service


playbooks/roles/certbot/tasks/main.yml
```diff
@@ -43,6 +43,14 @@
     path: /var/www/certbot
     state: directory
 
+- name: Create Certbot deploy hooks directory
+  file:
+    path: /etc/letsencrypt/renewal-hooks/deploy
+    state: directory
+
+- name: Copy the Certbot deploy hook to restart nginx on renewal
+  copy: src=restart-nginx.sh dest=/etc/letsencrypt/renewal-hooks/deploy/restart-nginx.sh mode=755
```
smarnach: Please use the multiline form here – we always prefer that style since we think it's more readable.

lgp171188:
@smarnach, I have modified the task to use the multi-line style.

playbooks/roles/certbot/tasks/main.yml
```diff
@@ -43,6 +43,17 @@
...
+- name: Create Certbot deploy hooks directory
+  file:
+    path: /etc/letsencrypt/renewal-hooks/deploy
+    state: directory
+
+- name: Copy the Certbot deploy hook to restart nginx on renewal
+  copy:
+    src: restart-nginx.sh
+    dest: /etc/letsencrypt/renewal-hooks/deploy/restart-nginx.sh
+    mode: 755
```
smarnach: This needs to be an octal number, so you need to prefix it with a zero.  (This is a rather unexpected pitfall.  When using the old style, it's automatically treated as octal by Ansible.  With the new style, the number is interpreted by the YAML parser instead of the Ansible code, so it's parsed according to the YAML rules.)

lgp171188: @smarnach, I wasn't aware of this behaviour till now. Will fix it.

lgp171188:
@smarnach, I have fixed the permission to use an octal value now. Will squash all the commits into one and then merge.

PR #65, opened 2018-07-23
Add a missing backslash.

UmanShahzad:
The mattermail config was incorrect without this. It is now fixed.

### DIFF
--- a/playbooks/roles/mattermail/defaults/main.yml
+++ b/playbooks/roles/mattermail/defaults/main.yml
@@ -16,4 +16,4 @@ MATTERMAIL_BRANCH: master
 MATTERMAIL_HOME: /opt/mattermail
 
 # Misc
-MATTERMAIL_MAIL_TEMPLATE: "@here :incoming_envelope: _From: **{{'{{'}}.From{{'}}'}}**_{{'\n'}}>_{{'{{'}}.Subject{{'}}'}}_"
+MATTERMAIL_MAIL_TEMPLATE: "@here :incoming_envelope: _From: **{{'{{'}}.From{{'}}'}}**_{{'\\n'}}>_{{'{{'}}.Subject{{'}}'}}_"


playbooks/roles/mattermail/defaults/main.yml
```diff
@@ -16,4 +16,4 @@ MATTERMAIL_BRANCH: master
 MATTERMAIL_HOME: /opt/mattermail
 
 # Misc
-MATTERMAIL_MAIL_TEMPLATE: "@here :incoming_envelope: _From: **{{'{{'}}.From{{'}}'}}**_{{'\n'}}>_{{'{{'}}.Subject{{'}}'}}_"
+MATTERMAIL_MAIL_TEMPLATE: "@here :incoming_envelope: _From: **{{'{{'}}.From{{'}}'}}**_{{'\\n'}}>_{{'{{'}}.Subject{{'}}'}}_"
```
smarnach: This looks like line noise to me.  How does this interact with Jinja2?  Anyway, I can see that the back slash is needed whatever it means. :)

smarnach: I realise now that `{{'{{'}}` is the Jinja2 way of escaping double braces.  This line would profit from adding the spaces we generally use: `{{ '{{' }}`.

smarnach: Or we could use

    {{ '_From: **{{.From}}**_\\n>_{{.Subject}}_' }}

to reduce the number of curly braces.  Still not the most readable, but maybe better than what we have.

UmanShahzad: @smarnach Oh jeez, that latter one is a lot better. I don't have much time to make a new PR and test just yet, but maybe in the next Mattermost iteration?

smarnach: Yeah, sorry for merging this permaturely – I'll try to think of it next months.

PR #68, opened 2018-08-12
Connect to non-replicated Redis deployment without Sentinel.

smarnach:
Since we are not going to use a highly available Redis deployment in the first iteration, I needed to adapt this role for a single-process Redis deployment.  For testing, it's enough to run the role against some app server.  End-to-end testing will be performed as part of [OC-4334](https://tasks.opencraft.com/browse/OC-4334).

### DIFF
--- a/playbooks/roles/appserver/defaults/main.yml
+++ b/playbooks/roles/appserver/defaults/main.yml
@@ -11,4 +11,4 @@ appserver_redis_port: 6379
 # Build upstream specs for the Consul Connect command line.
 appserver_memcached_upstream_specs: |-
   {% for port in appserver_memcached_ports %} -upstream memcached-{{ appserver_instance_id }}-{{ loop.index0 }}:{{ port }}{% endfor %}
-appserver_redis_upstream_specs: " -upstream redis-{{ appserver_instance_id }}-:{{ appserver_redis_port }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/appserver/defaults/main.yml
```diff
@@ -1,18 +1,14 @@
...
-  - 6380
-  - 6381
+appserver_redis_port: 6379
 
 # Build upstream specs for the Consul Connect command line.
 appserver_memcached_upstream_specs: |-
   {% for port in appserver_memcached_ports %} -upstream memcached-{{ appserver_instance_id }}-{{ loop.index0 }}:{{ port }}{% endfor %}
-appserver_redis_sentinel_upstream_specs: |-
-  {% for port in appserver_redis_sentinel_ports %} -upstream redis-sentinel-{{ appserver_instance_id }}-{{ loop.index0 }}:{{ port }}{% endfor %}
+appserver_redis_upstream_specs: " -upstream redis-{{ appserver_instance_id }}-:{{ appserver_redis_port }}"
```
UmanShahzad: Nit: There's a `-` before the colon and port. (In fact this is probably not a nit, as it may simply not work otherwise).

smarnach: I removed the job index, but failed to remove the dash.  Definitely not a nit! :)

PR #69, opened 2018-08-17
[OC-5066] Add playbook and roles for MIT annotation tools

viadanna:
This PR includes playbook and roles to deploy MIT annotation tools:
- [catchpy](https://github.com/nmaekawa/catchpy)
- [hxat](https://github.com/viadanna/hxat)
- [loris](https://github.com/loris-imageserver/loris)

The deployment can be tested on Vagrant using [this Vagrantfile](https://gist.github.com/viadanna/76a755d9f4f77db6506eed88f7cc6212) and this was already deployed to production servers.


playbooks/roles/hxat/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+hxat_django_allowed_hosts:
+  - 'localhost'
+  - '127.0.0.1'
+  - '*'
+
+hxat_django_loglevel: INFO
+hxat_django_secret: itsasecret
+
+hxat_organization: OpenCraft
```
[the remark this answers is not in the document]
viadanna: These are just the defaults for testing, I've made this change in [ansible-secrets](https://github.com/open-craft/ansible-secrets/blob/934d6488a8b42589b0fa51984e96a23a18fcb6ed/host_vars/mit-hxat.opencraft.hosting.yml#L10)

playbooks/roles/hxat/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
+# general
+
+hxat_app_repository: https://github.com/viadanna/hxat
```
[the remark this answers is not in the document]
viadanna: Good idea 👍

playbooks/roles/hxat/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+hxat_consumer_key: 123key
+hxat_lti_secret: secret
+hxat_consumer_url: null
+
+hxat_xframe_allowed_sites:
+  - canvas.localhost
+  - edx.localhost
+
+hxat_annotation_db_server: http://192.168.50.3
```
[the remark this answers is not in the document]
viadanna: Again, just the defaults for testing on Vagrant.

playbooks/roles/hxat/templates/secure.py.j2
```diff
@@ -0,0 +1,20 @@
+SECURE_SETTINGS = {
```
UmanShahzad: I have no clue what this PR is really about but I wanted to take a poke. Is there any reason these `.py` files aren't a part of the codebase, using a `.env` file to gather their settings from, where we only render the `.env` file through Ansible?

clemente: This secure.py file does the same role as the .env file, it concentrates all the custom settings in one place (well, almost all the settings we want to change). I [saw it used](https://github.com/harvard-vpal/bridge-adaptivity/blob/master/bridge_adaptivity/config/settings/secure_example.py) in another Harvard project.

UmanShahzad: We tend to follow 12-factor-app, which has this to say about configuration: https://12factor.net/config. This is the pattern we (and edX) use for basically everything config-wise, and is the norm in almost all of our deployments - which is why I find rendering full Python files for the settings to be a bit awkward to do.

Anyway, I'm not involved in this project, just a suggestion.

playbooks/roles/hxat/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 # Using our own repository to work around missing user names
 # This will become unnecessary when https://github.com/edx/xblock-lti-consumer/pull/32 is merged
 # hxat_app_repository: https://github.com/lduarte1991/hxat
-hxat_app_repository: https://github.com/viadanna/hxat
+hxat_app_repository: https://github.com/open-craft/hxat
 hxat_app_path: /srv/hxat
 hxat_version: v1.1.2.1
```
clemente: I think here you should use your branch in the new repo, `paulo/edx-lti`. You can leave the 1.1.2.1 as comment.

playbooks/roles/hxat/templates/secure.py.j2
clemente: Yes, I also don't like it and I pointed it out in the ticket. If we need to migrate often, we may need to improve this.
Now, we'd have to rewrite their code to read environment variables instead of their secure.py file, and contribute this upstream (to the Harvard project), and we didn't include this in the discovery.
Note that the this template inside the ansible role doesn't set hardcoded values, but it accesses ansible variables that are defined outside the role (e.g. in ansible-secrets).

playbooks/roles/hxat/defaults/main.yml
viadanna: 👍

playbooks/roles/hxat/templates/secure.py.j2
viadanna: Thanks for the suggestion, @UmanShahzad 
We discussed this awkward setup which is used by Harvard, and it would take more than a few changes to move away from it as mentioned by @clemente

clemente:
:+1: after the latest fixes.
The playbooks are probably not final because we'll be adding cache server, some way to upload files, etc. but after the client requests it.

Let's wait a little more until we know whether we need to add SSH access to loris.

clemente:
We can expand it for SSH in a later PR. SSH access isn't urgent.

clemente:
@viadanna Review after the new changes to update to 1.1.3:

I spawned a new server in OVH and ran the playbook. I verified that it doesn't fail and that it creates the files in the right places, and creates a `secure.py` with the right values (DEBUG=False, …) except for the two below. I could see the Django app work through an SSH tunnel, so it's working.

1. I saw this:
```
[Fri Sep 14 17:11:56.713914 2018] [wsgi:error] [pid 26788:tid 139740648773376] [client 127.0.0.1:44484]     dictConfigClass(config).configure()
[Fri Sep 14 17:11:56.713930 2018] [wsgi:error] [pid 26788:tid 139740648773376] [client 127.0.0.1:44484]   File "/usr/lib/python2.7/logging/config.py", line 576, in configure
[Fri Sep 14 17:11:56.713948 2018] [wsgi:error] [pid 26788:tid 139740648773376] [client 127.0.0.1:44484]     '%r: %s' % (name, e))
[Fri Sep 14 17:11:56.713967 2018] [wsgi:error] [pid 26788:tid 139740648773376] [client 127.0.0.1:44484] ValueError: Unable to configure handler 'default': [Errno 13] Permission denied: '/app.log'
```
It seems we need to add some way to end with this line in our secure.py:
```
    'log_root': '/srv/hxat',
```
(as we can see in the modified secure.py from Harvard). You can set it with the value of our ansible variable `hxat_app_path` which we already have (as default).


2. One reason to overwrite aws.py in the past was to change ALLOWED_HOSTS. Now this has improved and we just need to define  `allowed_hosts` in our secure.py (by taking the value from our ansible variable `hxat_django_allowed_hosts` that we already have); we must add it.

PR #70, opened 2018-08-25
Prometheus exporters

ahmedaljazzar:
There're two changes this PR is introducing to help achieving [OC-5001](https://tasks.opencraft.com/browse/OC-5001) goals:
- Changing the MongoDB exporter to help us collecting MongoDB query execution time.
- Enable collecting mysql query response time by default.

### DIFF
--- a/playbooks/roles/mongodb/defaults/main.yml
+++ b/playbooks/roles/mongodb/defaults/main.yml
@@ -8,9 +8,14 @@ MONGODB_SERVER_BACKUP_DIR: /var/lib/mongodb/backup
 # PROMETHEUS ##################################################################
 
 MONGODB_EXPORTER_ENABLED: true
-MONGODB_EXPORTER_VERSION: 1.0.0
-MONGODB_EXPORTER_DOWNLOAD_URL: https://github.com/dcu/mongodb_exporter/releases/download/v{{ MONGODB_EXPORTER_VERSION }}/mongodb_exporter-linux-amd64
+MONGODB_EXPORTER_VERSION: 0.6.1
+MONGODB_EXPORTER_REPO: https://github.com/percona/mongodb_exporter.git
 MONGODB_EXPORTER_PASSWORD: null
+
+MONGODB_HOME: /var/lib/mongodb
+GO_PATH: "{{ MONGODB_HOME }}"
+MONGODB_EXPORTER_DIR: "github.com/percona/mongodb_exporter"
+MONGODB_EXPORTER_PATH: "{{ GO_PATH }}/src/{{ MONGODB_EXPORTER_DIR }}"
 
 MONGODB_EXPORTER_CONSUL_SERVICE:
   service:
--- a/playbooks/roles/mongodb/defaults/main.yml
+++ b/playbooks/roles/mongodb/defaults/main.yml
@@ -8,12 +8,13 @@ MONGODB_SERVER_BACKUP_DIR: /var/lib/mongodb/backup
 # PROMETHEUS ##################################################################
 
 MONGODB_EXPORTER_ENABLED: true
-MONGODB_EXPORTER_VERSION: 0.6.1
+MONGODB_EXPORTER_VERSION: "v0.6.2"
 MONGODB_EXPORTER_REPO: https://github.com/percona/mongodb_exporter.git
 MONGODB_EXPORTER_PASSWORD: null
 
+# This is not meant to be changed, but to improve the readability and to reduce the repetition
 MONGODB_HOME: /var/lib/mongodb
-GO_PATH: "{{ MONGODB_HOME }}"
+GO_PATH: "/tmp/mongodb"
 MONGODB_EXPORTER_DIR: "github.com/percona/mongodb_exporter"
 MONGODB_EXPORTER_PATH: "{{ GO_PATH }}/src/{{ MONGODB_EXPORTER_DIR }}"
 
--- a/playbooks/roles/mongodb/tasks/main.yml
+++ b/playbooks/roles/mongodb/tasks/main.yml
@@ -72,7 +72,7 @@
   block:
     - name: Create custom fact directory
       file:
-        path: "/etc/ansible/facts.d"
+        path: /etc/ansible/facts.d
         state: "directory"
 
     - name: Store MongoDB exporter version
@@ -82,8 +82,8 @@
         dest: /etc/ansible/facts.d/mongodb_exporter.fact
       register: mongodb_exporter_version
 
-    - name: Upgrade MongoDB exporter
-      when: mongodb_exporter_version.changed
+    - name: Install and Upgrade MongoDB exporter
+      when: ansible_local.mongodb_exporter is not defined or ansible_local.mongodb_exporter.version != MONGODB_EXPORTER_VERSION
       block:
         - name: Install golang package
           apt:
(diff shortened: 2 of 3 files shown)


playbooks/roles/mongodb/defaults/main.yml
```diff
@@ -8,10 +8,15 @@ MONGODB_SERVER_BACKUP_DIR: /var/lib/mongodb/backup
 # PROMETHEUS ##################################################################
 
 MONGODB_EXPORTER_ENABLED: true
-MONGODB_EXPORTER_VERSION: 1.0.0
-MONGODB_EXPORTER_DOWNLOAD_URL: https://github.com/dcu/mongodb_exporter/releases/download/v{{ MONGODB_EXPORTER_VERSION }}/mongodb_exporter-linux-amd64
+MONGODB_EXPORTER_VERSION: 0.6.1
+MONGODB_EXPORTER_REPO: https://github.com/percona/mongodb_exporter.git
 MONGODB_EXPORTER_PASSWORD: null
 
+MONGODB_HOME: /var/lib/mongodb
```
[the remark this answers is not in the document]
ahmedaljazzar: So maybe I can write a comment saying that: This is not meant to be changed, but to improve the readability and to reduce the repetition". What do you think?

playbooks/roles/mongodb/defaults/main.yml
```diff
@@ -8,10 +8,15 @@ MONGODB_SERVER_BACKUP_DIR: /var/lib/mongodb/backup
...
+GO_PATH: "{{ MONGODB_HOME }}"
```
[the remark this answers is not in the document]
ahmedaljazzar: I don't know why I did such a stupid thing, actually! Sure, I'll just move it to a temp dir.

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -70,11 +70,33 @@
...
+        chdir: "{{ MONGODB_EXPORTER_PATH }}"
+        target: build
+      environment:
+        GOPATH: "{{ GO_PATH }}"
+
+    - name: create symlink to exporter binary
+      file:
+        src: "{{ MONGODB_EXPORTER_PATH }}/mongodb_exporter"
         dest: /usr/local/bin/mongodb_exporter
+        state: link
```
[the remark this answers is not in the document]
ahmedaljazzar: I struggled in this a lot tbh. I'll try doing it again and will let you know what's gonna happen.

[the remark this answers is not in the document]
smarnach: Not sure what exactly you struggled with.  For what it's worth, to copy files on the remote host you can use the `copy` module and set `remote_src: yes`.

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -70,12 +70,47 @@
...
-    - name: Download MongoDB exporter
-      get_url:
-        url: "{{ MONGODB_EXPORTER_DOWNLOAD_URL }}"
-        dest: /usr/local/bin/mongodb_exporter
-        mode: 0755
+    - name: Store MongoDB exporter version
+      copy:
+        content: |
+          {"version": "{{ MONGODB_EXPORTER_VERSION }}"}
+        dest: /etc/ansible/facts.d/mongodb_exporter.fact
```
smarnach: Shouldn't this file have a `.json` extension?

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -70,12 +70,47 @@
...
+      register: mongodb_exporter_version
+
+    - name: Upgrade MongoDB exporter
+      when: mongodb_exporter_version.changed
```
smarnach: The idea with local facts is that they are loaded automatically by Ansible.  You can access the fact as `ansible_local.mongodb_exporter.version`.

The problem with the current approach is that it is quite fragile.  The new version is written first, before the actual installation is performed.  If there is an error with some task during the installation, fixing the error and re-running the playbook won't have any effect – the fact file already contains the new version, so nothing will happen.

Instead, you should simply query `ansible_local.mongodb_exporter.version` before the installation.  It will be either unset, or contain the currently installed version.  If it is unset, or if the installed version differs from the desired version, we can proceed with the installation.  After the installation, we can update the fact file, so it only gets updated when the installation is successful.

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -91,11 +126,11 @@
         src: mongodb_exporter.service
         dest: /etc/systemd/system/mongodb_exporter.service
 
-    - name: Start MongoDB exporter
+    - name: Start and enable MongoDB exporter
       systemd:
         daemon-reload: yes
         name: mongodb_exporter.service
-        state: started
+        state: restarted
```
smarnach: The generally preferred solution is to use a handler for the restart, and only notify the handler if it's needed, but let's leave it as it is here, since it is good enough for now.

playbooks/roles/mongodb/tasks/main.yml
ahmedaljazzar: So as of Ansible's [setup - Gathers facts about remote hosts](https://docs.ansible.com/ansible/latest/modules/setup_module.html#parameters) documentation Ansible expects fact files used for local Ansible facts to be of `*.fact` extension, but the file/results format can be `json` or `ini-format`. I tried using `.json` extension but it didn't work so I changed it to be `.fact` instead.

playbooks/roles/mongodb/tasks/main.yml
ahmedaljazzar: Yeah I figured that out, but when I looked in the repo for similar examples I noticed that `restarted` is the common use, so I kept it like this. However, I'll try to change it while addressing the other comment.

playbooks/roles/mongodb/tasks/main.yml
ahmedaljazzar: Fair enough! However, trying to query `ansible_local.mongodb_exporter.version` before setting it up will cause a key error as Ansible not able to access the key from the fact. Will check how to overcome this and will implement the best practice.

smarnach: I think you should be able to use a condition like

    when: ansible_local.mongodb_exporter.version is not defined or ansible_local.mongodb_exporter.version != MONGODB_EXPORTER_VERSION

playbooks/roles/mongodb/tasks/main.yml
smarnach: This one isn't a big deal, so no need to spend much time on it.

ahmedaljazzar:
@smarnach This is ready for another pass.

playbooks/roles/mongodb/tasks/main.yml
```diff
@@ -70,12 +70,52 @@
...
-      get_url:
-        url: "{{ MONGODB_EXPORTER_DOWNLOAD_URL }}"
-        dest: /usr/local/bin/mongodb_exporter
-        mode: 0755
+    - name: Store MongoDB exporter version
+      copy:
+        content: |
+          {"version": "{{ MONGODB_EXPORTER_VERSION }}"}
+        dest: /etc/ansible/facts.d/mongodb_exporter.fact
+      register: mongodb_exporter_version
```
smarnach: The register isn't needed anymore.

PR #71, opened 2018-08-27
[BB-855] Install consul-template binary

UmanShahzad:
See https://github.com/open-craft/ansible-secrets/pull/103.

This only installs consul-template, and has a WIP version of lbv2, which I needed to add to actually test this somewhere live.

The lbv2 role will make use consul-template in ways described within this PR's discussion thread, in a future ticket.

### DIFF
--- a/playbooks/roles/load-balancer-v2/files/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/files/haproxy.cfg.ctmpl
@@ -1,0 +1,120 @@
+global
+    log /dev/log local0
+    log /dev/log local1 notice
+    chroot /var/lib/haproxy
+    stats socket /run/haproxy/admin.sock mode 660 level admin
+    stats timeout 2m
+    user haproxy
+    group haproxy
+    daemon
+
+    # Default SSL material locations
+    ca-base /etc/ssl/certs
+    crt-base /etc/ssl/private
+
+    # Default ciphers to use on SSL-enabled listening sockets.
+    # For more information, see ciphers(1SSL). This list is from:
+    #  https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
+    ssl-default-bind-ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS
+    ssl-default-bind-options no-sslv3
+
+    tune.ssl.default-dh-param 2048
+
+defaults
+    mode http
+    option http-server-close
+    option forwardfor
+    balance roundrobin
+    log global
+    option httplog
+    option dontlognull
+    option dontlog-normal
+    option redispatch
+    retries 3
+    timeout connect 5000
+    timeout client  50000
+    timeout server  2m
+    errorfile 400 /etc/haproxy/errors/400.http
+    errorfile 403 /etc/haproxy/errors/403.http
+    errorfile 408 /etc/haproxy/errors/408.http
+    errorfile 500 /etc/haproxy/errors/500.http
+    errorfile 502 /etc/haproxy/errors/502.http
+    errorfile 503 /etc/haproxy/errors/503.http
+    errorfile 504 /etc/haproxy/errors/504.http
+
+frontend fe_http
+    bind *:80
+    acl acme_challenge path_beg /.well-known/acme-challenge/
+    redirect scheme https code 301 unless acme_challenge
+    use_backend be_acme_challenge if acme_challenge
+
+frontend fe_https
+    bind *:443 ssl crt /etc/haproxy/certs.pem
+
+    # Use the error backend if the path begins with /error
+    acl error_mode path_beg /error
+    use_backend error if error_mode
+
+    # X-Forwarded-For is added by "option forwardfor"
+    http-request set-header X-Forwarded-Port %[dst_port]
+    http-request add-header X-Forwarded-Proto https
+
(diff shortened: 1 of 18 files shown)


smarnach:
@UmanShahzad Thanks for this!  This is looking good overall, but I haven't tested this yet.  There are still two main things I'm thinking about.

1. The old version of the load balancer had far two many moving parts that interact asynchronously and use `flock()` to avoid race conditions.  When reading the code again for writing the new version, I still found two race conditions that luckily didn't ever cause a problem.  For this reason, I worked on getting rid of the cert-watcher, and I hoped we would get rid of the haproxy-config-watcher and the cron job to reload haproxy as well.  I hoped the quiescence timers would simply cover rate limiting reloads.  If we set `wait.min` to one minute, we would at most reload once a minute, right?  If you agree that this works, I'd really prefer to get rid of the cron job.  Haproxy itself is wrapped by a wrapper binary that can be signalled to trigger a reload of haproxy, so we could simply configure consul-template to send that signal instead of `touch /tmp/haproxy-reload`.

2. We need some way to make sure the information in Consul is in a consistent state.  This should be a problem for _anyone_ using `consul-template`, so I wonder why there isn't any built-in locking mechanism.  The easiest solution probably is to grab a lock on the `ocim/instances` prefix before rendering the config files (or on `/certs` for the certificates).  We would need to add corresponding locks to Ocim and the cert manager, so they lock these prefixes before any modifications.

I suggest we wait until @ahmedaljazzar finished the actual templates before finishing this up.

UmanShahzad:
@smarnach I thought about consistent Consul reads through consul-template, but I'm not sure if it'll work so easily. We'd need to ensure consistency in Consul _after_ Consul Template's rate limit is taken into account, i.e. only when the daemon is about to render, and we can only do that within the consul template codebase itself. The inconsistent read will only happen if consul-template is reading, and Ocim suddenly does a write, right? With the quiscense timer, the case of consul-template doing a read while Ocim was already writing can be mostly eliminated.

So I think we'll need some wrapper around consul-template to make this work. Specifically, it should do this:

1. Do a Consul watch on key prefix `X/`
1. When key prefix `X/` changes, execute a Consul lock on `X/`
1. When the lock is taken, execute consul-template with `-once` and -wait=60s, so it waits 60 seconds then commits a render. Note that we need to use non-blocking reads in the templates (using just "key" to fetch a key is blocking), otherwise we could be in a deadlock situation if some key doesn't exist.
1. Once consul-template executes, the lock is automatically dropped, and we go back into a watching state.

On the Ocim/cert manager side, we need to just implement locks before doing any write.

UmanShahzad:
We also need multiple watchers with this scheme unfortunately, which means there will potentially be more than 1 consul-template renderings happening at once if the prefixes are different, e.g. `ocim/instances` and `certs/` happening at the same time. That leads back to the issue of doing multiple reloads at the same time.

UmanShahzad:
For that reason I think we should put "certs" into "ocim/certs", and then just watch "ocim/".

playbooks/roles/vault/tasks/main.yml
```diff
@@ -38,7 +38,7 @@
   notify: restart vault
 
 - name: Determine current Vault version
-  shell: vault version | sed -n 's/Vault v\(.*\)/\1/p'
+  shell: vault version | cut -d" " -f 2 - | cut -b 2- -
```
UmanShahzad: Misc fix - I was doing it wrong in a previous PR, and this fixes it. Used same tactic for consul-template version detection, so decided to utilize it here.

playbooks/roles/load-balancer-v2/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+    - haproxy.cfg.ctmpl
+
+- name: Copy over consul-template configuration files
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ consul_template_config_dir }}/{{ item }}"
+  with_items:
+    - backend.map.hcl
+    - certs.pem.hcl
+    - haproxy.cfg.hcl
```
SSPJ: I didn't see these files on haproxy-c-1.net.opencraft.hosting -- I guess this is because you didn't run this role on that host yet, but I am noting it just in case.

playbooks/roles/load-balancer-v2/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+
+- name: Ensure that {{ maintenance_server_static_root }} exists
```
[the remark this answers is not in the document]
UmanShahzad: It actually does work if the variable doesn't change depending on the inventory - I'm doing as much in my private ansible roles. But yeah I need to double-check this - will do it in a future PR when we finish up the lbv2 role.

playbooks/roles/load-balancer-v2/files/haproxy.cfg.ctmpl
```diff
@@ -0,0 +1,120 @@
...
+
+# When the AppServer raises the "sinking ship" 50x error this backend gets
+# activated and serves an alternate error page that can be customised
+# per-instance.
+backend error
+    server error-page 127.0.0.1:{{ maintenance_server_port }}
+
+# The following backends can be manually activated during times of scheduled
+# or unscheduled maintenance. The instructions for how to activate them can be
+# found here: https://github.com/open-craft/documentation/blob/master/something_is_down.md
```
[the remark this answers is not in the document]
UmanShahzad: Ah thanks, it should be https://gitlab.com/opencraft/documentation/private/blob/master/ops/incidents.md.

playbooks/roles/load-balancer-v2/tasks/main.yml
UmanShahzad: Yep, didn't run this, as it's not complete - just used the tag.

PR #72, opened 2018-08-28
[OC-5141] Invert logic in tarsnap role.

UmanShahzad:
The current version is only configuring tarsnap when the tarsnap key isn't specified or is empty, which is the opposite of what it should be doing.

### DIFF
--- a/playbooks/roles/tarsnap/tasks/main.yml
+++ b/playbooks/roles/tarsnap/tasks/main.yml
@@ -1,11 +1,10 @@
 ---
 
 - include: install.yml
-  tags:
-    - tarsnap
+  tags: tarsnap
+
 - include: configure.yml
-  tags:
-    - tarsnap
   tarsnap_cache: "{{ TARSNAP_CACHE }}"
   tarsnap_keyfile: "{{ TARSNAP_KEY_REMOTE_LOCATION }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/tarsnap/tasks/main.yml
```diff
@@ -1,11 +1,10 @@
...
-    - tarsnap
+  tags: tarsnap
+
 - include: configure.yml
-  tags:
-    - tarsnap
   tarsnap_cache: "{{ TARSNAP_CACHE }}"
   tarsnap_keyfile: "{{ TARSNAP_KEY_REMOTE_LOCATION }}"
-  when: "TARSNAP_KEY is not defined or TARSNAP_KEY == ''"
+  when: "not (TARSNAP_KEY is not defined or TARSNAP_KEY == '')"
```
lgp171188: @UmanShahzad, wouldn't it be better to drop the negation and use `TARSNAP_KEY is defined and TARSNAP_KEY != ''`. Makes it easier to read and understand.

UmanShahzad: Yep definitely better.

UmanShahzad:
@lgp171188 Can I get an explicit :+1:  here btw?

UmanShahzad:
@lgp171188 Thanks!

PR #75, opened 2018-09-14
Various fixes for Mattermost and the autolinker.

smarnach:
This fixes a few issues with the Mattermost server:

* Mattermail will be killed if it uses too much memory, so it does not take Mattermost down.
* The Mattermost plugin directory is created if it does not exist yet.
* We now use the new `mattermost` binary instead of the deprecated `platform` binary.
* Correcly handle links with a query string in the autolinker.

### DIFF
--- a/playbooks/roles/mattermail/templates/mattermail.service.j2
+++ b/playbooks/roles/mattermail/templates/mattermail.service.j2
@@ -11,6 +11,9 @@ Restart=always
 RestartSec=10
 User=mattermail
 Group=mattermail
+# Mattermail seems to be leaking memory, so we let systemd kill it if it uses too much memory.
+MemoryHigh=120M
+MemoryMax=150M
 
 [Install]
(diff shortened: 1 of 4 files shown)


playbooks/roles/mattermail/templates/mattermail.service.j2
```diff
@@ -11,6 +11,9 @@ Restart=always
 RestartSec=10
 User=mattermail
 Group=mattermail
+# Mattermail seems to be leaking memory, so we let systemd kill it if it uses too much memory.
+MemoryHigh=120M
+MemoryMax=150M
```
UmanShahzad: Just to confirm, it'll get restarted automatically by systemd right, or will it just sit dead in case it got OOM killed?

UmanShahzad: Sounds good.

PR #81, opened 2018-10-20
Collect top metrics in MongoDB exporter

ahmedaljazzar:


playbooks/roles/mongodb/files/mongodb_exporter.service.j2
```diff
@@ -3,7 +3,7 @@ Description=Prometheus MongoDB Exporter
 After=network-online.target
 
 [Service]
-ExecStart=/usr/local/bin/mongodb_exporter -web.listen-address 127.0.0.1:9001
+ExecStart=/usr/local/bin/mongodb_exporter {% if MONGODB_EXPORTER_EXTRA_OPTS != "" %}{{ MONGODB_EXPORTER_EXTRA_OPTS }}{% endif %} -web.listen-address 127.0.0.1:9001
```
smarnach: The `if` seems redundant.  In case `MONGODB_EXPORTER_EXTRA_OPTS == ""`, including it would inlcude the empty string, so just remooving the `if` results in the same output.

PR #84, opened 2018-11-26
[SE-535] Set up new user management SFTP server

SSPJ:
All this does is make sure sftp is enabled, which it is by default in Ubuntu.

I've changed the Subsystem command to be `internal-sftp` rather than `sftp-server` as it is now. That should make no difference.

My first thought was that we should have this chrooted, but on further thought I concluded that this is a dedicated server. The only thing of value on it is the sftp files which a chroot user would have access to anyway. So I am leaving it as-is.


SSPJ:
@smarnach I moved stuff to `playbooks/roles/common-server-init/tasks/sftp.yml` and tested it on a vagrant box. Let me know if it looks good to merge.

SSPJ:
@smarnach I put the flag back, defaulting to false, and tested on a vagrant box. As sftp is enabled by default in openssh on [Ubuntu 16.04](https://launchpad.net/ubuntu/+source/openssh/1:7.2p2-4), it only makes a difference if we explicitly disable it elsewhere. :)

smarnach:
@SSPJ I didn't know it's enabled by default – it wasn't the case on some Ubuntu release I used in the past (or maybe it was Debian, don't quite remember). So we wouldn't really have needed this, right? Anyway, it's better to explicitly enable the setting.

PR #85, opened 2018-12-12
WIP: [SE-683] Nexus learner dashboard role

smarnach:


playbooks/roles/notre-dame-nexus-dashboard/templates/ecosystem.yml.j2
```diff
@@ -0,0 +1,44 @@
...
+    env       :
+      NODE_ENV    : 'staging'
+      APP_URI     : 'http://{{ nexus_dashboard_host }}:{{ nexus_dashboard_port }}/'
+      MC_HOST     : '0.0.0.0'
+      MC_PORT     : '{{ nexus_dashboard_port }}'
+      DB_MAIN_URI : 'mongodb://{{ nexus_dashboard_mongo_host }}:{{ nexus_dashboard_mongo_port }}/ndplatform'
+      CH_MAIN_URL : 'redis://{{ nexus_dashboard_redis_host }}:{{ nexus_dashboard_redis_port }}'
+      WE_IDP_CS   : '{{ nexus_dashboard_idp_client_secret }}'
+      WE_IDP_RU   : 'http://{{ nexus_dashboard_host }}:{{ nexus_dashboard_port }}/auth/idp/callback'
+      WR_GOOGLE_CS: none
```
smarnach: I think we need a variable to set this, right? Similarly for many of the other variables here.

Another option to render this file is to create a dictionary in Ansible first, e.g.

    nexus_dashboard_configuration:
      apps:
        - name: ndp
          [...]
        - name: ndw
          [...]

For rendering the file, you can then use

    - name: Create ecosystem.yml file
      copy:
        content: "{{ nexus_dashboard_configuration | to_nice_yaml }}"
        dest: /notredame/apps/ecosystem/ecosystem.yml

This would avoid all the boilerplate for the configuration variables – we would basically write the whole configuration file in the Ansible variables instead.

playbooks/roles/notre-dame-nexus-dashboard/templates/staging.js.j2
```diff
@@ -0,0 +1,70 @@
...
+  },
+  agenda: {
+    jobs: {
+      edxCourseSync: { recurrence: '75 minutes' },
+      googleEventSync: { recurrence: '45 minutes' },
+      rssFeedUpdate: { recurrence: '10 minutes' }
+    }
+  }
+};
+
```
smarnach: Nit: trailing whitespace

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -128,17 +126,16 @@
         "EnableSignInWithUsername": true,
         "SendEmailNotifications": true,
         "RequireEmailVerification": false,
-        "FeedbackName": "Mattermost Notification",
+        "FeedbackName": "ND MM Notification",
```
UmanShahzad: I might be misunderstanding something, but we use the mattermost role for chat.opencraft.com too

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -302,9 +384,70 @@
...
     "DataRetentionSettings": {
-        "Enable": false
+        {% if {{ MATTERMOST_VERSION }} == '4.0.0' %}
+            "Enable": false
+        {% else %}
+            "EnableMessageDeletion": false,
+            "EnableFileDeletion": false,
+            "MessageRetentionDays": 365,
+            "FileRetentionDays": 365,
+            "DeletionJobStartTime": "02:00"
```
SSPJ: Jinja2 templates aren't white space sensitive like Python. `{% if %}` blocks must be ended with `{% endif %}`.

You should have gotten a parsing error when you tested this, but if not it definitely isn't going to work the way you expect.

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -127,32 +186,37 @@
...
-        "PushNotificationServer": "",
+        "PushNotificationServer": "{{ MATTERMOST_PUSH_NOTIFICATION_SERVER }}",
         "PushNotificationContents": "generic",
-        "EnableEmailBatching": false,
+        "EnableEmailBatching": true,
         "EmailBatchingBufferSize": 256,
         "EmailBatchingInterval": 30,
         "SkipServerCertificateVerification": false
+        {% if {{ MATTERMOST_VERSION }} != '4.0.0' %}
+            "SkipServerCertificateVerification": false,
```
SSPJ: It doesn't make sense to test on MATTERMOST_VERSION unless what you are testing for is version dependent. Since I don't see any tests based on host I am guess you are using version as a stand in for host? It would be better to use inventory host directly. 

This `SkipServerCertificateVerification` line occurs inside and outside the block.

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -90,46 +98,59 @@
...
         "EnableFile": true,
         "FileLevel": "INFO",
-        "FileJson": true,
+        {% if {{ MATTERMOST_VERSION }} == '4.0.0' %}
+            "FileFormat": "",
+        {% else %}
+            "FileJson": true,
         "FileLocation": "{{ MATTERMOST_LOG_DIR }}",
-        "EnableWebhookDebugging": false,
+        "EnableWebhookDebugging": true,
```
SSPJ: This line isn't protected by a variable.

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -164,38 +186,37 @@
         "EnableSignInWithEmail": true,
         "EnableSignInWithUsername": true,
         "SendEmailNotifications": true,
-        "UseChannelInEmailNotifications": false,
```
SSPJ: This line seems to have been removed?

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -164,38 +186,37 @@
...
-        "RequireEmailVerification": true,
-        "FeedbackName": "Mattermost Notification",
+        "RequireEmailVerification": {{ MATTERMOST_REQUIRE_EMAIL_VERIFICATION }},
+        "FeedbackName": "{{ MATTERMOST_FEEDBACK_NAME }}",
         "FeedbackEmail": "{{ MATTERMOST_FEEDBACK_EMAIL }}",
         "FeedbackOrganization": "",
-        "EnableSMTPAuth": false,
```
SSPJ: Ditto for missing line.

playbooks/roles/mattermost/templates/config.json.j2
```diff
@@ -164,38 +186,37 @@
...
-        "SMTPUsername": "",
-        "SMTPPassword": "",
-        "SMTPServer": "localhost",
-        "SMTPPort": "25",
-        "ConnectionSecurity": "",
+        "SMTPUsername": "{{ MATTERMOST_SMTP_USERNAME }}",
+        "SMTPPassword": "{{ MATTERMOST_SMTP_PASSWORD }}",
+        "SMTPServer": "{{ MATTERMOST_SMTP_SERVER }}",
+        "SMTPPort": "{{ MATTERMOST_SMTP_PORT }}",
+        "ConnectionSecurity": "TLS",
```
SSPJ: `ConnectionSecurity` should probably also be behind a variable.

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: I went over it more carefully and corrected it. I also ran the playbook to ensure the config file gets parsed correctly.

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: I'm not entirely sure on what basis the config differs i.e. whether it depends only on version or something else. I figured testing based on version would make more sense since we use an older version and so it's possible some of the features are not supported. Also none of the other services using the same playbook are using it (default version was set to 5). 
Are you suggesting testing against particular hosts? i.e. we'll have a check like `if {{ MATTERMOST_HOSTNAME }} == 'x.x.x.x'`. That value will need to be changed for each host right? Or is there another way to do that?

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: Modified it to use a variable.

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: Thank you. I missed it. I've added it now.

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: Missed this too. Added it back.

playbooks/roles/mattermost/templates/config.json.j2
pkulkark: Done.

bradenmacdonald:
@pkulkark @SSPJ What's the plan for this PR?

SSPJ:
Hi @bradenmacdonald , thanks for checking in.

This PR needs to be merged eventually, because we want to use the ansible-playbooks to deploy Notre Dame's custom dashboard app. At present, it is WIP because there are conflicts with the ND mattermost deployment and OpenCraft's own mattermost deployment which are being resolved by inserting ansible variables.

However, we don't have enough budget from ND to finish it so this will probably remain open for some months.

bradenmacdonald:
I see. Yeah the `{% if MATTERMOST_VERSION == '4.0.0' -%}` approach definitely needs to be changed to use host variables before we could merge this. Thanks for the info. BTW could you please add the JIRA ticket number to the PR header or description?

SSPJ:
@bradenmacdonald Yep, done.

smarnach:
Since there hasn't been any activity here for three years, I'm closing this (it always shows up in the list of my open PRs).

bradenmacdonald:
Good call. Thanks and hi @smarnach 👋🏻  :)

PR #86, opened 2019-01-02
Playbook for setting up an Elastic Stack server

symbolist:
Blockstore uses ElasticSearch 6. This is a role to configure a server to act as a dedicated ElasticSearch server. Currently it is not production ready and may need more configuration in the future.

- Uses the official Ansible role for installing Elastic Stack 6.
- ElasticSearch is behind nginx with basic auth configured.


SSPJ:
:+1: Good work finding the right galaxy role for ElasticSearch, @symbolist

PR #87, opened 2019-01-03
[BB-11] Add Wordpress

Agrendalath:
This PR adds playbooks for deploying *.opencraft.com sites.


playbooks/roles/common/files/hosts
```diff
@@ -0,0 +1,28 @@
+127.0.0.1 localhost
```
lgp171188: @Agrendalath, we may not need most of `plebia.org` entries since we only use the OpenCraft domains for almost all the production stuff.

playbooks/roles/nginx/files/sites/cardstories/cardstories.org.conf
```diff
@@ -0,0 +1,70 @@
+server {
```
lgp171188: @Agrendalath, is this needed any more? I am not sure if it is used. Also, the cardstories.org site uses an expired SSL certificate

playbooks/roles/nginx/files/sites/mail/00_default_mail.plebia.org.conf
```diff
@@ -0,0 +1,41 @@
+server {
```
lgp171188: @Agrendalath, I don't think this is needed either or did I miss something?

playbooks/roles/nginx/files/sites/media/media.plebia.org.conf
```diff
@@ -0,0 +1,42 @@
+server {
```
lgp171188: @Agrendalath, I think we should keep all the non-OpenCraft stuff out of this repository.

playbooks/roles/nginx/files/sites/opencraft-console/console.opencraft.com.conf
```diff
@@ -0,0 +1,57 @@
+server {
```
lgp171188: @Agrendalath, this is confusing. Don't we have a separate configuration set up and in use for OCIM?

playbooks/roles/nginx/files/sites/opencraft-tasks/tasks.opencraft.com.conf
```diff
@@ -0,0 +1,60 @@
+server {
```
lgp171188: @Agrendalath, the `tasks.opencraft.com` domain is pointed to a different VM. So we should not include this in this PR. Ditto for other OpenCraft sites not hosted on the old opencraft.com VM.

.gitmodules
```diff
@@ -20,3 +20,15 @@
 	path = playbooks/roles/load-balancer
 	url = https://github.com/open-craft/ansible-load-balancer
 	branch = master
+[submodule "playbooks/roles/geerlingguy.nginx"]
```
lgp171188: @Agrendalath, can we not use the existing `nginx` role in this repository and avoid using a separate, duplicate sub-module?

playbooks/roles/common/files/etckeeper.conf
```diff
@@ -0,0 +1,18 @@
+export GIT_AUTHOR_NAME="etckeeper"
```
lgp171188: @Agrendalath, this is covered in the `common-server-init` role already.

playbooks/roles/common/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+- name: Copy hosts file
```
lgp171188: @Agrendalath, it looks like most, if not all of this is already covered in the `common-server-init` role. Can we just reuse that and only add whatever is missing in that?

playbooks/roles/wordpress/files/sites/default.opencraft.com.conf
```diff
@@ -0,0 +1,31 @@
+server {
```
lgp171188: @Agrendalath, all the host-specific configuration should be moved to the `ansible-secrets` repository in a file under the `host_vars` directory and only the generic playbook files and templates should be in this repository.

playbooks/roles/wordpress/tasks/nginx.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: chown www-data /var/www
+  file: path=/var/www mode=750 owner=www-data group=www-data state=directory
+
+- name: mkdir /var/www/log
+  file: path=/var/www/log mode=750 state=directory
+
+- name: Copy nginx global configuration
+  synchronize: src="global/" dest=/etc/nginx/global checksum=yes delete=yes owner=no group=no
+
+- name: Copy nginx site configurations
```
lgp171188: @Agrendalath, hardcoding the site name is not a good idea. This will have to be made generic using variables which can be overridden for specific hosts or sites.

playbooks/roles/wordpress/tasks/nginx.yml
```diff
@@ -0,0 +1,32 @@
+---
```
lgp171188: @Agrendalath, don't you think a lot of this is covered by the existing `nginx-proxy` role? Is there a reason why this should be added separately?

playbooks/wordpress.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Deploy the OpenCraft.com server with wordpress
+  hosts: wordpress
+  become: true
+  vars:
+    - OPS_EMAIL: "ops@example.com"
```
lgp171188: @Agrendalath, this is fine as long as we override it in the host-specific configuration file with the correct email address. But it might be better to remove all the variable definitions and move them to that file.

playbooks/wordpress.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Deploy the OpenCraft.com server with wordpress
```
lgp171188: @Agrendalath, this has to be generic and not specific to one particular site.

playbooks/roles/wordpress/files/sites/default.opencraft.com.conf
Agrendalath: Done.

playbooks/roles/wordpress/tasks/nginx.yml
Agrendalath: Done.

playbooks/wordpress.yml
Agrendalath: That's right. I removed it.

playbooks/roles/wordpress/tasks/nginx.yml
Agrendalath: We are running them on the same port, so I'm not sure what we would gain by setting up an `nginx-proxy` here. Am I missing something?

playbooks/wordpress.yml
Agrendalath: Done.

playbooks/roles/wordpress/tasks/nginx.yml
lgp171188: @Agrendalath, nginx acts as a proxy in this case because the actual Wordpress code is run on the php-fpm servers. So my question was about using the `nginx-proxy` role to do all the below tasks instead of repeating it all here.

playbooks/wordpress.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Deploy the OpenCraft landing page
+  hosts: wordpress
+  become: true
+  vars:
+    - nginx_remove_default_vhost: true
+    - php_memory_ilimit: "128M"
+    - php_max_execution_time: "90"
+    - php_upload_max_filesize: "256M"
+    - php_version: "5.6"
```
lgp171188: @Agrendalath, is there any reason for using the `5.6` version which is no longer supported when Ubuntu 16.04 has the 7.0 version?

playbooks/roles/wordpress/templates/wordpress-nginx.j2
```diff
@@ -0,0 +1,87 @@
...
+        location /files {
+                autoindex off;
+                autoindex_exact_size off;
+                autoindex_localtime on; 
+
+                #auth_basic "Restricted";
+                #auth_basic_user_file /var/www/.htpasswd_opencraftstaff;
+        }   
+
+        # Wordpress config
```
lgp171188: @Agrendalath, can we keep the global include configuration in `global/wordpress.conf` like it is done now so that it is reusable?

playbooks/roles/wordpress/templates/default-nginx.j2
```diff
@@ -0,0 +1,31 @@
...
+		try_files $uri $uri/ =404;
+	}
+
+	location ~ [^/]\.php(/|$) {
+		fastcgi_split_path_info ^(.+?\.php)(/.*)$;
+		if (!-f $document_root$fastcgi_script_name) {
+			return 404;
+		}
+
+		fastcgi_pass unix:/var/run/php5-fpm.sock;
```
lgp171188: @Agrendalath, it might be a good idea to use create and separate FPM pools for each site.

playbooks/roles/wordpress/README.md
```diff
@@ -0,0 +1,10 @@
+Ansible role for OpenCraft's landing page
+============================================================
+
+This role sets up a Wordpress-based landing page, maintenance page and default page. The service is run behind nginx, and
+certbot is used to manage SSL certificates for the landing page.
+
+After deploying new landing page, you need to:
+- mount volume with Wordpress data (it can be created from a snapshot),
+- create new MySQL DB from the dump and set it in `wp-config.php`,
```
lgp171188: @Agrendalath, can this be automated if the path to the local database dump is provided?

playbooks/roles/wordpress/README.md
```diff
@@ -0,0 +1,10 @@
...
+- change hostname of the Worpdress to the correct one in the new DB with `UPDATE opncrft_options SET option_value='https://{{ HOST }}' WHERE option_name = 'siteurl' OR option_name = 'home';`.
```
lgp171188:
```suggestion
- change hostname of the Wordpress to the correct one in the new DB with `UPDATE opncrft_options SET option_value='https://{{ HOST }}' WHERE option_name = 'siteurl' OR option_name = 'home';`.
```

playbooks/roles/wordpress/templates/default-nginx.j2
Agrendalath: Hm, I don't think we actually need it here. The `default` page doesn't have any PHP files. I'll just remove it.

playbooks/roles/wordpress/README.md
Agrendalath: Ah, good catch :D

playbooks/wordpress.yml
Agrendalath: I don't know if the current version of Wordpress and its plugins are fully compatible with it. If we updated it, we would need to check if everything is working correctly there and I don't have enough context on what might break here. So it's possible, but would require more time to test it properly.

playbooks/roles/wordpress/templates/wordpress-nginx.j2
Agrendalath: Sure :+1:

playbooks/roles/wordpress/README.md
Agrendalath: Sure, but we should do a follow-up for this as it significantly changes the current scope.

playbooks/wordpress.yml
lgp171188: @Agrendalath, I run Wordpress sites using PHP 7 and they work fine. But it is out of scope for this PR and should be looked at in a different task. We might ditch PHP for the v3 version of the website.

playbooks/roles/wordpress/tasks/nginx.yml
Agrendalath: Yep, we are using it as a proxy, but we also host files with it, so it's not only the proxy. We could reuse this role, but we would need to overwrite its template to be able to use it for our needs (which would greatly decrease its readability) or overwrite the role itself to allow custom templates (and risk breaking some other roles that depend on this one).

lgp171188: @Agrendalath, we should leave it as-is and leave this improvement and the investigation for it, for a later task

playbooks/roles/wordpress/tasks/nginx.yml
```diff
@@ -0,0 +1,44 @@
...
+
+- name: Find needed nginx configurations
+  find:
+    paths: /etc/nginx/sites-available/wordpress
+  register: found
+
+- name: Enable nginx site configurations
+  file:
+    src: "{{ item.path }}"
+    dest: /etc/nginx/sites-enabled/{{ item.path | basename }}
```
lgp171188: @Agrendalath, can we name the configuration files `<site domain name>.conf`?

Agrendalath: Yup, that makes sense. Done.

playbooks/roles/wordpress/tasks/nginx.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- name: Add landing site nginx configuration
+  template:
+    src: ../templates/wordpress-nginx.j2
+    dest: /etc/nginx/sites-available/wordpress/{{ LANDING_SITE_HOSTNAME }}.conf
+
+- name: Add default site nginx configuration
+  template:
+    src: ../templates/default-nginx.j2
+    dest: /etc/nginx/sites-available/wordpress/{{ DEFAULT_SITE_HOSTNAME }}.conf
```
lgp171188: @Agrendalath, nit: does this default site have to be under the `wordpress` sub-directory since it has nothing to do with Wordpress. Ditto for the maintenance site.

playbooks/roles/wordpress/tasks/nginx.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- name: Ensure directory for config files exists
+  file:
+    path: /etc/nginx/sites-available/wordpress
+    state: directory
+
+- name: Copy nginx global configuration
+  synchronize: src="global/" dest=/etc/nginx/global checksum=yes delete=yes owner=no group=no
+
+- name: Add landing site nginx configuration
```
lgp171188: @Agrendalath, nit: might be a good idea to combine this with the next 2 tasks by using placeholders and the `with_items`.

playbooks/roles/wordpress/tasks/nginx.yml
Agrendalath: Done.

playbooks/roles/wordpress/tasks/nginx.yml
Agrendalath: Done.

PR #92, opened 2019-01-31
Minor hxat upgrade, to 1.1.5

clemente:
Differences: https://github.com/lduarte1991/hxat/compare/v1.1.4..v1.1.5

Solves a problem with a popup, and a bug when answering to a comment which had a deleted reply.


clemente:
Merging because it's a minor change and they're reviewing this functionality in their server.

PR #97, opened 2019-02-14
[BB-741] add molecule tests to vault

always-boycott-github:
This is so I can write tests for the Vault SSH role.

I realize this is not the usual way of test, which is to run against the staging instance. If you would rather not have molecule tests, I'm fine giving it up. It helped me understand the playbooks, learning by doing ;-)

If you like it, I could add a .travis.yml so it acts as a kind of CI.

```bash
$ molecule -s vault test
...
--> Executing Testinfra tests found in /home/loic/software/opencraft/ansible-playbooks/molecule/vault/tests/...
    ============================= test session starts ==============================
    platform linux -- Python 3.6.6, pytest-4.2.1, py-1.7.0, pluggy-0.8.1 -- /home/loic/software/opencraft/virtualenv-ansible-playbook/bin/python3
    rootdir: /home/loic/software/opencraft/ansible-playbooks/molecule/vault, inifile:
    plugins: testinfra-1.14.0
collected 1 item                                                               
    
    tests/test_vault.py::test_vault[ansible://vault-host] PASSED             [100%]
    
    =========================== 1 passed in 3.01 seconds ===========================
...
```


playbooks/roles/vault/templates/config.hcl.j2
```diff
@@ -14,10 +14,12 @@ listener "tcp" {
     tls_require_and_verify_client_cert = "{{ "true" if vault_tcp_tls_require_and_verify_client_cert else "false" }}"
 }
 
+{% if vault_consul_acl_token is defined %}
```
UmanShahzad: Tokens are optional, so although internally we do specify it, we should probably use something like `vault_consul_storage: <bool>` for deciding to render this or not instead.

playbooks/roles/vault/tasks/main.yml
```diff
@@ -9,7 +9,6 @@
   user:
     name: vault
     group: vault
-    createhome: no
```
UmanShahzad: Is this intentional? This will change behavior because the default is `yes` - https://docs.ansible.com/ansible/2.5/modules/user_module.html.

molecule/vault/molecule.yml
```diff
@@ -0,0 +1,43 @@
...
+    - destroy
+    - create
+    - converge
+    - verify
+    - destroy
+verifier:
+  name: testinfra
+  options:
+    v: True
+#    s: True
```
UmanShahzad: Stray comment?

molecule/vault/group_vars/vault/public.yml
```diff
@@ -0,0 +1,111 @@
...
+#        L                      = Test Locality
+#        O                      = Organization Name
+#        OU                     = Organizational Unit Name
+#        CN                     = 127.0.0.1
+#        emailAddress           = test@email.address
+#
+#        [ req_attributes ]
+#        challengePassword              = A challenge password
+#
+# openssl req -config /tmp/l '-nodes' '-new' '-x509' '-keyout' 'server.key' '-out' 'server.crt'
```
UmanShahzad: Just for your interest, I used https://github.com/cloudflare/cfssl to generate all the certs. I've felt that the config files it uses are a bit more friendly than OpenSSL's.

always-boycott-github: thanks for the tip!

always-boycott-github:
> Adding TravisCI integration is probably not necessary but if you'd like to you can.

I will abstain then, it's out of scope.

playbooks/roles/vault/tasks/main.yml
always-boycott-github: Yes, this is intentional. Without it vault -dev will fail to create ~/.vault-token.

playbooks/roles/vault/templates/config.hcl.j2
always-boycott-github: I'll do that: better be explicit about it. :+1:

molecule/vault/molecule.yml
always-boycott-github: I like to keep it around for debugging: it's easier to remove the **#** than type again the **s: True**. But it's from a time when I had trouble remembering how molecule maps options to testinfra :-) I'll remove it.

always-boycott-github:
Addressed the comments, rebased and repushed. @UmanShahzad if that helps, I like to maintain a clean commit history and reading each commit in order may provide a view of the proposed changes that is less hairy than the global diff.

playbooks/roles/vault/tasks/main.yml
UmanShahzad: Technically it makes no difference, but it's generally unneeded. It might be confusing though because other roles don't create homes, so to be consistent we probably shouldn't create it, and you can change this with a flag.

UmanShahzad:
@always-boycott-github Yeah I generally agree with leaving around the full commit history.. I think it's mostly upstream that we squash because edX has explicitly asked us to do that for a long time.

UmanShahzad:
@always-boycott-github Btw I'm sure no one has used Molecule before, so if you could just leave a super small README.md in the molecule/ folder explaining why this is here, that'd be good. Then you can merge this.

playbooks/roles/vault/tasks/main.yml
always-boycott-github: I'll do that to be consistent, good point.

always-boycott-github:
README.md added. Thanks for the quick review @UmanShahzad !

playbooks/roles/vault/tasks/main.yml
```diff
@@ -61,7 +61,7 @@
 
 - name: Install Vault environment variables
   lineinfile:
-    path: "/home/{{ ansible_ssh_user }}/.bashrc"
+    path: "{{ ansible_env.HOME }}/.bashrc"
```
pilou-: Note that `~` or `$HOME` could be used instead of `ansible_env.HOME`, because [type of `path` parameter is `path`](https://github.com/ansible/ansible/blob/3408de156ab089e9f05c912d52869fee20fa5fc8/lib/ansible/modules/files/lineinfile.py#L34) and [`path` type parameters automatically expand `~`/`$HOME`](https://github.com/ansible/ansible/blob/3408de156ab089e9f05c912d52869fee20fa5fc8/lib/ansible/module_utils/basic.py#L1949-L1951).

molecule/vault/tests/test_vault.py
```diff
@@ -0,0 +1,6 @@
+def test_vault(host):
+    cmd = host.run("bash -c 'PS1=yes ; source /root/.bashrc ; vault status'")
+    print(cmd.stdout)
+    print(cmd.stderr)
+    assert 'Initialized     true' in cmd.stdout
```
pilou-: `vault status -format=json` could be used (parsing `json` output is easier than parsing default `table` format).

README.md
```diff
@@ -245,3 +245,8 @@ This ansible repository deploys MySQL server on OpenStack provider.
 
 1. Check that backups are saved to tarsnap
 1. Check contents of these backups.
+
+### Run molecule tests
+
+1. pip install -r test-requirements.txt
+1. molecule -s vault test
```
pilou-: `s/molecule -s vault test/molecule test -s vault/`

molecule/vault/molecule.yml
```diff
@@ -0,0 +1,42 @@
...
+driver:
+  name: docker
+lint:
+  name: yamllint
+platforms:
+  # docker + systemd recipe courtesy https://github.com/pilou-
+  - name: vault-host
+    image: xenial
+    command: /sbin/init
+    security_opts: ['seccomp=unconfined']
```
pilou-: Note that [Fedora seccomp profile](https://src.fedoraproject.org/rpms/docker/raw/master/f/seccomp.json) could be used.

test-requirements.txt
```diff
@@ -0,0 +1,3 @@
+-r requirements.txt
+molecule==2.16.0
+docker-py-1.10.6
```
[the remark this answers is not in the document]
always-boycott-github: Yeah, that mistake was fixed in a later PR. Good catch :-)

playbooks/roles/vault/tasks/main.yml
always-boycott-github: Excellent, it's less arcane :+1:

README.md
always-boycott-github: This was fixed in another PR, stupid mistake on my part and good catch :-)

molecule/vault/molecule.yml
always-boycott-github: I should really learn more about apparmor. It is also needed in the context of codejail, a library used by Open edX which requires specific apparmor settings to be tested.

always-boycott-github: In which context was this Fedora seccomp created and why is it a good fit in this case?

molecule/vault/tests/test_vault.py
always-boycott-github: I will do that in [another, larger, pull request](https://github.com/open-craft/ansible-playbooks/pull/99/files#diff-1be18bdbd3a11bbc7711b8c27df13bdaR7)

always-boycott-github:
@pilou- thanks a lot for this post-merge review :smiley:

molecule/vault/molecule.yml
pilou-: This Fedora seccomp profile whitelists system calls required by Systemd.

PR #99, opened 2019-02-15
[BB-741] vault ssh roles

dupainaulevain:
The vault-ssh.yml playbook is intentionally limited to a single host, backup-pruner, so the vault-ssh-{signer,host} roles can be applied in production with a limited scope. Additional hosts and more users can be added later, following this simple example.


playbooks/vault-ssh.yml
```diff
@@ -0,0 +1,20 @@
...
+
+  tasks:
+    - include_role:
+        name: vault-ssh-signer
+      vars:
+        vault_ssh_signer_ca: "{{ item }}"
+      with_items:
+        - backup-pruner
+
+- name: Dessiminate vault ssh CA
```
UmanShahzad: Typo: disseminate

playbooks/roles/vault-ssh-signer/defaults/main.yml
```diff
@@ -0,0 +1,16 @@
...
+vault_ssh_signer_ca: ca
+
+vault_config_dir: "/etc/vault"
+vault_tls_dir: "{{ vault_config_dir }}/tls"
+vault_tcp_tls_cert_file: "{{ vault_tls_dir }}/cert.pem"
+vault_tcp_tls_key_file: "{{ vault_tls_dir }}/key.pem"
+vault_tcp_tls_client_ca_file: "{{ vault_tls_dir }}/ca.pem"
+
+vault_env:
+    VAULT_TOKEN: "{{ vault_policy_mgmt_token }}"
```
UmanShahzad: The policy management token can only has full capabilities on `sys/policies/*`. I think we'll need some "ssh-signer-manager" policy or something, whose token we could use here. Otherwise this role won't run properly.

playbooks/roles/vault-ssh-signer/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+        "allowed_users": "*",
+        "default_extensions": [
+          {
+            "permit-pty": ""
+          }
+        ],
+        "key_type": "ca",
+        "default_user": "ubuntu",
+        "ttl": "720h0m0s"
+      }
```
UmanShahzad: Possible for this whole JSON to be a variable? Might make it easier to reference / override where necessary later.

playbooks/roles/vault-ssh-host/defaults/main.yml
```diff
@@ -0,0 +1,20 @@
+---
+vault_policy_mgmt_token: testtoken
+vault_ssh_signer_ca: ca
+
+vault_host: vault-a-1.net.opencraft.hosting
```
UmanShahzad: This should be set in ansible-secrets.

playbooks/roles/vault-ssh-host/defaults/main.yml
```diff
@@ -0,0 +1,20 @@
...
+vault_server_scheme: http
+
+vault_config_dir: "/etc/vault"
+vault_tls_dir: "{{ vault_config_dir }}/tls"
+vault_tcp_tls_cert_file: "{{ vault_tls_dir }}/cert.pem"
+vault_tcp_tls_key_file: "{{ vault_tls_dir }}/key.pem"
+vault_tcp_tls_client_ca_file: "{{ vault_tls_dir }}/ca.pem"
+
+vault_env:
+    VAULT_TOKEN: "{{ vault_policy_mgmt_token }}"
```
UmanShahzad: Similar comment about this token, though we can probably use a less capable policy for this role.

playbooks/vault-ssh.yml
always-boycott-github: Nice piece of borken English indeed :-P

playbooks/roles/vault-ssh-signer/defaults/main.yml
always-boycott-github: Oh, indeed. I incorrectly assumed it was a root token, my bad. I will change that to `vault_ssh_signer_mgmt_token` and test accordingly.

playbooks/roles/vault-ssh-signer/tasks/main.yml
always-boycott-github: Good idea, will do :+1:

playbooks/roles/vault-ssh-host/defaults/main.yml
always-boycott-github: ok, will do

playbooks/roles/vault-ssh-signer/defaults/main.yml
dupainaulevain: done and this also  led to move shared variables in the vault group

playbooks/roles/vault-ssh-host/defaults/main.yml
dupainaulevain: it turns out to be needlessly complicated. Instead the CA are stored in files and retrieved when needed.

playbooks/roles/vault-ssh-host/defaults/main.yml
dupainaulevain: done

playbooks/roles/vault-ssh-signer/tasks/main.yml
dupainaulevain: done

PR #100, opened 2019-02-18
portability fixes for molecule

always-boycott-github:
Based on suggestions from @pilou- It can be tested with

- pip install -r test-requirements.txt
- molecule test -s vault

### DIFF
--- a/playbooks/roles/vault/tasks/main.yml
+++ b/playbooks/roles/vault/tasks/main.yml
@@ -61,7 +61,7 @@
 
 - name: Install Vault environment variables
   lineinfile:
-    path: "{{ ansible_env.HOME }}/.bashrc"
+    path: "$HOME/.bashrc"
     line: "export {{ item }}"
   with_items: "{{ vault_env }}"
 
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,3 @@
 -r requirements.txt
-molecule==2.16.0
-docker-py==1.10.6
+molecule==2.19.0
+docker==3.7.0


playbooks/roles/vault/tasks/main.yml
```diff
@@ -61,7 +61,7 @@
 
 - name: Install Vault environment variables
   lineinfile:
-    path: "{{ ansible_env.HOME }}/.bashrc"
+    path: "$HOME/.bashrc"
```
UmanShahzad: So it looks like both of these actually spawn the changes into /root/.bashrc, whereas I thought they'd go into /home/ubuntu/.bashrc. I actually didn't know it'd go there in the first place, but I think that it's probably alright to go to root by default... although in practice it makes no difference because we aren't controlling security on those users so strictly.

UmanShahzad:
Argh I merged before actually trying `molecule test -s vault` :) I get this:

```
--> Validating schema /<redacted>/ansible-secrets/deploy/molecule/vault/molecule.yml.
Validation completed successfully.
--> Test matrix
    
└── vault
    ├── destroy
    ├── create
    ├── converge
    ├── verify
    └── destroy
    
--> Inventory /<redacted>/ansible-secrets/deploy/molecule/vault/group_vars linked to /tmp/molecule/deploy/vault/group_vars
--> Inventory /<redacted>/ansible-secrets/deploy/molecule/vault/host_vars linked to /tmp/molecule/deploy/vault/host_vars
--> Scenario: 'vault'
--> Action: 'destroy'
    
    PLAY [Destroy] *****************************************************************
    
    TASK [Destroy molecule instance(s)] ********************************************
    changed: [localhost] => (item=None)
    
    TASK [Wait for instance(s) deletion to complete] *******************************
    failed: [localhost] (item=None) => {"attempts": 1, "censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
    
    PLAY RECAP *********************************************************************
    localhost                  : ok=1    changed=1    unreachable=0    failed=1   
    
    
ERROR: 
An error occurred during the test sequence action: 'destroy'. Cleaning up.
--> Inventory /<redacted>/ansible-secrets/deploy/molecule/vault/group_vars linked to /tmp/molecule/deploy/vault/group_vars
--> Inventory /<redacted>/ansible-secrets/deploy/molecule/vault/host_vars linked to /tmp/molecule/deploy/vault/host_vars
--> Scenario: 'vault'
--> Action: 'destroy'
    
    PLAY [Destroy] *****************************************************************
    
    TASK [Destroy molecule instance(s)] ********************************************
    changed: [localhost] => (item=None)
    
    TASK [Wait for instance(s) deletion to complete] *******************************
    failed: [localhost] (item=None) => {"attempts": 1, "censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
    
    PLAY RECAP *********************************************************************
    localhost                  : ok=1    changed=1    unreachable=0    failed=1   
    
    
ERROR: 
```

UmanShahzad:
FYI it doesn't happen with the old test-requirements.txt.

playbooks/roles/vault/tasks/main.yml
always-boycott-github: Good catch. Since the goal here is to ensure the logged-in user has the vault variables set, it would be a regression for people who login as the ubuntu user, right? I entirely missed that the first time around.

UmanShahzad: Yeah it would be but one can easily `sudo -s` -> `source /root/.bashrc`. But yeah, it'd be nicer if we could just get it on the ubuntu user, since that's where all the other stuff is.

always-boycott-github:
I'm not able to reproduce the problem. The following works fine for me:

- virtualenv --python=python3 ../virtualenv-ansible-playbooks
-  source ../virtualenv-ansible-playbooks/bin/activate
- pip install -r test-requirements.txt
- molecule --debug test -s vault

as of b6ab5017384c6ec0713f764db7af46743f0e1223. If the problem persists. The --debug will help figure out what environmental issue is causing the test to fail on your local machine. Very sorry for the trouble :-(

playbooks/roles/vault/tasks/main.yml
always-boycott-github: I'll fix it then

UmanShahzad:
@always-boycott-github Hmm, must've been something wrong on my side then, because it suddenly works now without any changes... :) Thanks!

PR #101, opened 2019-02-19
vault environment is for ubuntu, not root

always-boycott-github:
also split the vault-env.sh out of .bashrc so it can conveniently be
included


always-boycott-github:
@UmanShahzad this should fix the regression you had [here](https://github.com/open-craft/ansible-playbooks/pull/100#discussion_r257869478)

PR #102, opened 2019-02-19
vault environment is for ubuntu, not root

always-boycott-github:
@UmanShahzad this should fix the regression you had [here](https://github.com/open-craft/ansible-playbooks/pull/100#discussion_r257869478)

### DIFF
--- a/playbooks/roles/vault/tasks/main.yml
+++ b/playbooks/roles/vault/tasks/main.yml
@@ -18,7 +18,7 @@
     state: directory
     owner: "{{ item.owner }}"
     group: vault
-    mode: 0750
+    mode: 0755
   with_items:
     - { path: "{{ vault_bin_dir }}", owner: root }
     - { path: "{{ vault_log_dir }}", owner: vault }
--- a/playbooks/roles/vault/tasks/main.yml
+++ b/playbooks/roles/vault/tasks/main.yml
@@ -54,16 +54,22 @@
   notify: restart vault
 
 - name: Configure `vault` autocomplete
-  shell: "vault -autocomplete-install"
   become: no
+  shell: "vault -autocomplete-install"
   when: vault_binary.changed
   ignore_errors: yes
 
 - name: Install Vault environment variables
+  become: no
+  template:
+    src: vault-env.j2
+    dest: "/home/{{ ansible_ssh_user }}/.vault-env"
+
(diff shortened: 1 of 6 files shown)


UmanShahzad:
Thanks @always-boycott-github, will check this out soon.

playbooks/roles/vault/tasks/main.yml
```diff
@@ -54,16 +54,22 @@
...
+    dest: "/home/{{ ansible_ssh_user }}/.vault-env"
+
+- name: Source Vault environment variables
+  become: no
   lineinfile:
-    path: "$HOME/.bashrc"
-    line: "export {{ item }}"
-  with_items: "{{ vault_env }}"
+    path: "/home/{{ ansible_ssh_user }}/.bashrc"
+    line: "source $HOME/.vault-env"
```
UmanShahzad: I guess this is where Ansible's lack of convergence fails us - now `.bashrc` has both the old `export X=Y` lines plus this new one! It isn't harmful, but it's less than desirable. For now I'll just manually go in and delete those lines.

dupainaulevain: Yeah. I should have provided a cleanup playbook to remove the spurious lines. I'm still very fuzzy regarding when and how the playbooks are applied. Do you do that manually from time to time? With tags? Maybe there is a document I could read explaining this?

dupainaulevain:
Thanks for the review & the merge :tada:  It's really nice to see things moving!

playbooks/roles/vault/tasks/main.yml
UmanShahzad: Yep it's super manual :p I just run it whenever I need to deploy something. I wish we ran them on a cron for all servers regularly, but the truth is some playbooks are _not_ necessarily completely idempotent, in particular some of the external ones. So it's dangerous to run them automatically, but I think long-term that'd be a really important improvement.

I do run with tags, and when I need to be surgical I use `--step`. I used `--step` here to ensure I didn't accidentally restart Vault by accident. Doing that would mean we'd have to unlock it with the unlock keys, and that'll get dirty :)

I'm not sure about docs, I think the beginning part of https://github.com/open-craft/ansible-secrets/ mostly explains it - just need to run `ansible-playbook` with whatever parameters are correct for your goal.

PR #104, opened 2019-03-03
[BB-741] revert vault ssh

dupainaulevain:


UmanShahzad:
@Agrendalath I'll have to let you review this. I feel like there's some missing stuff here, but you may have better context.

dupainaulevain:
I thought about what could be missing. But I think it's a clean revert: it removes two roles as well as the playbooks from which they are used and no other playbook or role depends on them.

PR #108, opened 2019-03-22
[BB-840] Fix the when condition to not use jinja2 delimiters and apt module warning

lgp171188:
Fixes the below warning

```
[WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: "{{ caching_resolver_enabled | default(false) }}"
```

and the warning from the `apt` module that the `installed` state is deprecated and that the `present` state should be used.

### DIFF
--- a/playbooks/load-balancer.yml
+++ b/playbooks/load-balancer.yml
@@ -15,5 +15,5 @@
       tags: 'node-exporter'
 
     - role: unbound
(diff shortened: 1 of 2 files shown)


lgp171188:
@giovannicimolin, thanks for the quick review on short notice! :+1:

PR #114, opened 2019-05-20
[BB-1237] Render Ocim instances' SSL certs into separate files from consul

lgp171188:
This fixes the issue of HAProxy serving the wrong certificate for its domains as HAProxy expects to have just one certificate in each PEM file.

There is now a bash-script-as-a-service which watches the `ocim/certs` prefix, acquires the lock and then writes all the individual certificates (one file per instance) after deleting all the existing certificates. To make this easier, all the SSL certificates for Ocim instances have been moved into a separate sub-directory.

The HAProxy configuration has been updated to load the certificates from `/etc/haproxy/certs`, the directory which is used currently, and `/etc/haproxy/certs/ocim` which is the new sub-directory containing all the SSl certificates for Ocim instances.

**Testing instructions**
* Deploy the changes using the playbooks from this branch.
* Check and verify that the certificate files are written properly to the correct file under the `/etc/haproxy/certs/ocim` directory.
* Also verify that HAProxy is able to load all the certificates without any errors and serves the correct certificate for all the domains.
* Also verify that the service updates the certificates properly when the certificate data in consul changes.


playbooks/roles/load-balancer-v2/defaults/main.yml
```diff
@@ -7,8 +7,10 @@ haproxy_domain_regex: "haproxy.*\\.net\\.opencraft\\.hosting"
 haproxy_ppa: "ppa:vbernat/haproxy-1.8"
 haproxy_version: "1.8.*"
 
-haproxy_consul_watch_prefix: "ocim/"
+haproxy_consul_watch_prefix: "ocim/instances"
```
SSPJ: I recall there was a reason that the `haproxy_consul_watch_prefix` had to be `ocim/` rather than `ocim/instances` but can't remember the details. I will have to dig that info up to make sure this is okay.

lgp171188: @SSPJ, sure. I made this change because Uman had mentioned in that comment that this had to be done. Otherwise I have no prior context on this.

lgp171188: @SSPJ, did you find out about why the watch prefix had to be `ocim/` instead of `ocim/instances` as recommended by Uman?

PR #115, opened 2019-05-21
BB-1269: Update opencraft website deployment scripts [WIP]

giovannicimolin:
Testing instructions on https://github.com/open-craft/ansible-secrets/pull/120

### DIFF
--- a/playbooks/roles/wordpress/tasks/wordpress.yml
+++ b/playbooks/roles/wordpress/tasks/wordpress.yml
@@ -8,85 +8,14 @@
     - unzip
     - mysql-client
     - build-essential
-    - python-dev
+    - python3-dev
     - libmysqlclient-dev
-    - python-mysqldb
-    - python-pip
+    - python3-mysqldb
+    - python3-pip
 
 - name: Install the Python MySQLB module
   pip:
     name: mysqlclient
-
-- name: Ensure .ssh directory exists for www-data user
-  file:
-    path: "/var/www/.ssh/"
-    state: directory
-    owner: www-data
-    group: www-data
-
-- name: Create key file
-  no_log: True
-  copy:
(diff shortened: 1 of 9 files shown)


playbooks/roles/wordpress/tasks/wordpress.yml
```diff
@@ -0,0 +1,126 @@
...
+    - python-dev
+    - libmysqlclient-dev
+    - python-mysqldb
+    - python-pip
+
+- name: Install the Python MySQLB module
+  pip:
+    name: mysqlclient
+
+- name: Ensure .ssh directory exists for www-data user
```
lgp171188: @giovannicimolin, wouldn't it be better to do this all as a root/sudo user and then just setting the permissions to the `www-data` user at the end?

playbooks/wordpress.yml
```diff
@@ -7,21 +7,21 @@
...
-      - php-curl
-      - php-imap
-      - php-json
-    - php_mysql_package: php5.6-mysql
+      - php7.3-gd
+      - php7.3-curl
+      - php7.3-imap
+      - php7.3-json
+    - php_mysql_package: php7.3-mysql
   roles:
```
lgp171188: @giovannicimolin, we don't have the backups enabled for the VM this playbooks runs on. You will have to set and specify the `TARSNAP_BACKUP_FOLDERS`  variable to get it to work. You should backup at least the `/var/www`, `/etc/nginx/sites-available` directories and also create a snitch and configure it in the `ansible-secrets` playbook after configuring all the `TARSNAP_*` variables.

giovannicimolin: @lgp171188 Setting up backups is not in the scope of this ticket.

PR #117, opened 2019-06-05
[SE-994] Remove websockets configuration

clemente:
Due to swampdragon being deprecated, we're removing it from Ocim until we can find a good replacement.
To keep the change consistent, we're removing the websocket configuration variables here, and we'll readd them when websockets are reimplemented.
This includes:
- the instructions to point a domain like `websockets.console.opencraft.com` to Ocim. We're not deleting the domain itself
- the nginx proxy that makes it listen in port 443 and redirects it to port 2001
- the certbot config that will get SSL certificates for the websockets domain
- in addition I'm deleting some old variables that aren't used anywhere (ocim, roles here, configuration, …) and that should have been deleted at https://github.com/open-craft/ansible-playbooks/commit/d15701e1a69f656a97e15336a3a157a42d9e4810

*Testing instructions*

- similar to https://github.com/open-craft/opencraft/pull/433
- after that PR works and is deployed, deploy these changes to Ocim stage (without merging the PR yet) and check that it removes the nginx config (or do it manually if ansible doesn't want to remove configs)
- things should still be working same as before, in terms of websockets (should already been tested in the previous PR)
- If things work, merge the PR then apply the changes to Ocim production too

*Notes*
This will require redeploying Ocim by running the ansible playbook.


clemente:
Checked by @swalladge, thanks.

In addition I just deployed Ocim stage (ansible-secrets, ansible-playbooks) and manually removed the nginx configuration, and restarted it, and it works.

  - [X] I tested it
  - [X] I read through the code
  - [X] I checked for accessibility issues
  - [X] Includes documentation: I commented out the websocket config in Ocim stage and linked to this ticket.

I'll deploy to Ocim production now.

PR #120, opened 2019-07-15
 [BB-1247] Add Sprints

Agrendalath:
This is related to open-craft/sprints#6.

### DIFF
--- a/playbooks/roles/sprints/tasks/main.yml
+++ b/playbooks/roles/sprints/tasks/main.yml
@@ -1,4 +1,5 @@
-- name: Stop Sprints (this is reuqired for `common-sever` role to pass)
+# This is required for `common-sever` role to pass, because Let's Encrypt uses Nginx on port 80, which is being used by this container.
+- name: Stop Sprints
   docker_compose:
     project_src: "{{ APP_PATH }}"
     stopped: yes
@@ -44,6 +45,7 @@
   tags:
     - initial
 
+# This config is created by `common-server` role for Let's Encrypt and is no longer needed here.
 - name: Remove unused nginx config
   file:
     path: /etc/nginx/sites-enabled/common-server
--- a/playbooks/roles/sprints/tasks/main.yml
+++ b/playbooks/roles/sprints/tasks/main.yml
@@ -1,0 +1,43 @@
+- name: Set authorized key for CircleCI
+  authorized_key:
+    user: ubuntu
+    state: present
+    key: "{{ CI_SSH_PUBLIC_KEY }}"
+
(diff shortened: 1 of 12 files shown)


playbooks/roles/sprints/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
+- name: Stop Sprints (this is reuqired for `common-sever` role to pass)
```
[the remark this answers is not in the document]
Agrendalath: Good catch - that was a bit confusing. I added the explanation in the comment.

Agrendalath:
@giovannicimolin, please take a look at [this template](https://github.com/open-craft/ansible-playbooks/pull/120/files#diff-7f79de24f08ccd480fd14243aecabf8b) and at our config - we're not hosting the Postgres server on the same machine as the VM :wink: 

But, answering your questions:
> Do we have any important data stored?

Nah, at the moment we're just storing social auth credentials and user accounts (not containing any specific data) in the DB.


> Is it hard to recover it from the backups if we experience a failure?

If we decided to host Postgres on the same VM for some reason, we [would be](https://github.com/open-craft/ansible-playbooks/pull/120/files#diff-7f79de24f08ccd480fd14243aecabf8bR64) storing the backups in the S3.

playbooks/roles/sprints/tasks/main.yml
```diff
@@ -44,6 +45,7 @@
   tags:
     - initial
 
+# This config is created by `common-server` role for Let's Encrypt and is no longer needed here.
 - name: Remove unused nginx config
```
giovannicimolin: I just noticed this now, but isn't this used when renewing the certificates?
I think that the common-role sets this up to renewal (Let's Encrypt certificates expire every 3 months).
If the sprints container uses port 80, how can we renew the certificates automatically?

Agrendalath: For SSL, [we're using](https://github.com/open-craft/sprints/blob/master/compose/production/traefik/traefik.toml) built-in Let's Encrypt support in Traefik (like described [here](https://docs.traefik.io/user-guide/docker-and-lets-encrypt/)).

I didn't find any easy way to disable Let's Encrypt in `common-sever` role, so I didn't tinker around this role to avoid breaking all other playbooks.

giovannicimolin: @Agrendalath I think that node exporter uses an https endpoint that prometheus uses to fetch data. I'm not really sure about this now, but it can be checked again later, when the certificate is near it's expiry date.

PR #121, opened 2019-07-17
BB-1466: Reduce curator verbosity to reduce spam

giovannicimolin:
This PR sets a few configuration flags to reduce the amount of spam sent by curator.

**Reviewer: **
- [ ] @xirdneh

### DIFF
--- a/playbooks/roles/curator/templates/action.yml.j2
+++ b/playbooks/roles/curator/templates/action.yml.j2
@@ -5,6 +5,7 @@ actions:
     description: "Delete indices older than {{ curator_delete_indices_days }} days."
     options:
       ignore_empty_list: True
+      continue_if_exception: True
     filters:
       - filtertype: kibana
         exclude: true
--- a/playbooks/roles/curator/templates/action.yml.j2
+++ b/playbooks/roles/curator/templates/action.yml.j2
@@ -6,6 +6,7 @@ actions:
     options:
       ignore_empty_list: True
       continue_if_exception: True
+      timeout_override: 300
     filters:
       - filtertype: kibana
         exclude: true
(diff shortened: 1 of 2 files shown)


playbooks/roles/curator/templates/action.yml.j2
```diff
@@ -3,6 +3,8 @@ actions:
   1:
     action: delete_indices
     description: "Delete indices older than {{ curator_delete_indices_days }} days."
+    options:
+      ignore_empty_list: True
```
xirdneh: Should we add `continue_if_exception: True` in case there's an error with one of the indices we can continue and delete the rest instead of waiting for somebody to fix the issue?
Also, I'm not sure which indices are we deleting here but I'm think that's probably old indices that hold edx courses metadata for search. I believe those indices can grow quite big, should we also add a way to manage the timeout in case there's an issue in the future?
`timeout_override: <seconds>`

giovannicimolin: @xirdneh These are the logs for almost everything from our infrastructure. We only keep them for 6 months because it's a massive amount of data and we don't have enough storage to keep that around for more. Also, it doesn't make much sense to keep logs for more time than that.
Does `continue_if_exception` logs the error? Because if it doesn't it might lead to the logs machine to fail silently due to out of disk space errors, so that's why I'd rather not include it.

xirdneh: @giovannicimolin I would expect that weather the exception gets logged or not has to do with the logging configuration, but I'm not entirely sure about it.
Given the size of the indices I think it would be a good idea to see if we can add these two config values since it's recommended for very large indices.

giovannicimolin:
@xirdneh I tested and added the `timeout_override` option. 
I will merge this now, thanks for the review!

PR #129, opened 2019-08-08
[SE-1376] Consul and cert manager for CI tests

SSPJ:
This PR prepares the load-balancer-v2 role for deploying the CI load balancer by adding the appropriate group vars and updating the role.

It is WIP and will do the same for the CI cert-manager in the next set of commits.


playbooks/group_vars/load-balancer-v2-ci/public.yml
```diff
@@ -0,0 +1,50 @@
+---
+
+# GENERAL #####################################################################
+
+LOAD_BALANCER_SERVER_IP: "{{ COMMON_FLOATING_IP or ansible_host }}"
+cert_manager_server_address: "51.38.226.128"
```
giovannicimolin: This is a public repo, can you move this IP address to the `ansible-secrets` repo?

PR #133, opened 2019-09-06
[BB-1533] Set up metrics exporting for Sprints

Agrendalath:
This enables exporting metrics from the sprints dashboard. It removes traefik (which collided with our standard playbooks) and uses `nginx-proxy` for redirecting the traffic directly to the Gunicorn exposed to the server's localhost from the Docker container. It also enables and starts `nodeexporter.service`, which was disabled and stopped by default for some reason.


playbooks/roles/sprints/tasks/main.yml
```diff
@@ -1,20 +1,26 @@
...
+  ufw:
+    rule: allow
+    port: "{{ item }}"
+    proto: tcp
+  with_items:
+    - 80
+    - 443
+
+# For some reason this service wasn't started after running playbooks.
+- name: Enable and start node-exporter systemd service to start on boot automatically
```
giovannicimolin: This is probably a bug on ansible somewhere. I'm seeing this happen on a few instances, but the upstream role that we use to deploy node exporter enables the service (or at least tries to). See https://github.com/ernestas-poskus/ansible-prometheus/blob/4d30cd59a60f50cc23bffc071bb8f99131f2dd8f/handlers/main.yml#L17.

giovannicimolin: You can leave this code here :+1:

Agrendalath: @giovannicimolin, it seems to be a difference between using [systemd_module](https://docs.ansible.com/ansible/latest/modules/systemd_module.html) and [service_module](https://docs.ansible.com/ansible/latest/modules/service_module.html). We're using 18.04 for this instance, so it might be related to the removal of Upstart after Ubuntu 14.04, but it's just my guess.

PR #137, opened 2019-09-27
[SE-888] Fixes for the mattermost autolink plugin

lgp171188:
This PR contains the fixes for deploying the mattermost auto-link plugin and was tested in the most-recent upgrade to 5.15.0

**Testing instructions**:
* Upgrade Mattermost using the changes in this PR and the corresponding changes in the related ansible secrets PR.

### DIFF
--- a/playbooks/roles/mattermost/defaults/main.yml
+++ b/playbooks/roles/mattermost/defaults/main.yml
@@ -5,10 +5,6 @@ MATTERMOST_OPS_EMAIL: ops@exmaple.com
 
 MATTERMOST_VERSION: 5.8.0
 
-# UNLINK is the autolink plugin which converts JIRA tickets to clickable URL
-# https://github.com/mattermost/mattermost-plugin-autolink/releases/
-MATTERMOST_UNLINK_VERSION: 0.3.0
-
 # Restrict registration to email addresses in this domain
 MATTERMOST_RESTRICT_TO_EMAIL_DOMAIN: example.com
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/mattermost/defaults/main.yml
```diff
@@ -4,6 +4,9 @@ MATTERMOST_HOSTNAME: "{{ inventory_hostname }}"
 MATTERMOST_OPS_EMAIL: ops@exmaple.com
 
 MATTERMOST_VERSION: 5.8.0
+
+# UNLINK is the autolink plugin which converts JIRA tickets to clickable URL
+# https://github.com/mattermost/mattermost-plugin-autolink/releases/
 MATTERMOST_UNLINK_VERSION: 0.3.0
```
SSPJ: ~~Can we update the default to `1.1.0` also?~~

Actually, can we remove this default and move the comment to the secrets repo? I'm not a big fan of defaults for vars that don't really have any meaningful universal fall-back value.

In the case of versions, that's either "latest" or, for some software packages, a well-recognized standard value, such as "2.7" or "3.6". </soapbox> :)

playbooks/roles/mattermost/tasks/mattermost.yml
```diff
@@ -32,10 +32,14 @@
     group: mattermost
     force: no
 
+# This autolink configuration is NOT used UNLESS you are installing MM from scratch.
+# You must MANUALLY copy it from /var/lib/mattermost/autolink.json to the correct place
+# in /var/lib/mattermost/config.json and restart MM.
```
SSPJ: Suggestion: "This is because the playbook never updates config.json on an existing installation."

SSPJ:
I left a couple of nits, @lgp171188 but otherwise this looks good to merge once the submodule is updated. :+1:

- [x] I tested this: similar code for installing `5.13.0`
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ n/a
- [ ] ~~Includes documentation~~ n/a

playbooks/roles/mattermost/defaults/main.yml
lgp171188: Done.

PR #138, opened 2019-10-02
BB-1667 Update chkrootkit expected output daily

viadanna:
Description
=========

This PR adds a daily cron entry that updates the expected `chkrootkit` output with whatever was the last output.

This results in each change to the output being reported only once to ops.

Testing
----------

1. Deploy this to a VM which runs the chkrootkit daily.
2. If the vm wasn't notifying the chkrootkit diff, make sure to modify the chkrootkit expected log to make it notify.
3. Check there's no difference between the chkrootkit expected and today logs and that the vm stops notifying after one day.

### DIFF
--- a/playbooks/roles/common-server-init/templates/update-chkrootkit-log
+++ b/playbooks/roles/common-server-init/templates/update-chkrootkit-log
@@ -6,4 +6,7 @@
 
 LOG_DIR=/var/log/chkrootkit
 
-cp -a -f $LOG_DIR/log.today $LOG_DIR/log.expected
+if ! diff -q $LOG_DIR/log.expected $LOG_DIR/log.today > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


playbooks/roles/common-server-init/templates/update-chkrootkit-log
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+#
+# This script updates the expected log daily so each change is notified only
+# once.
+#
+
+LOG_DIR=/var/log/chkrootkit
+
+cp -a -f $LOG_DIR/log.today $LOG_DIR/log.expected
```
lgp171188: @viadanna, it might be a good idea to retain the version before the current version of the expected file.

viadanna: :+1:

PR #141, opened 2019-10-22
BB-1667 Fix src script

viadanna:
This fixes the script name for `update-chkrootkit-log`.

### Testing

1. Deploy any role that contains the `common-server-init` role.
1. Make sure it does not fail when copying the script.

```
$ ansible-playbook -l localhost playbooks/ocim.yml
(...)
TASK [common-server-init : Update expected output daily] ***********************
changed: [localhost]
(...)
```

### DIFF
--- a/playbooks/roles/common-server-init/tasks/chkrootkit.yml
+++ b/playbooks/roles/common-server-init/tasks/chkrootkit.yml
@@ -27,7 +27,7 @@
 
 - name: Update expected output daily
   template:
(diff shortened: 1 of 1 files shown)


lgp171188:
@viadanna, after merging this PR, we have to update the submodule hash in the `ansible-secrets` repository. I will take care of that while doing so for the Ocim websockets work.

viadanna:
Thanks @lgp171188

PR #142, opened 2019-10-29
Prevent corrupting HAProxy cfg with missing IPs

SSPJ:
These are a few additions to the Consul template which renders the HAProxy configuration from the values which Ocim stores in the `ocim/instances/` prefix.

Specifically, if Ocim sets the IP address to "nil" (this should never happen and yet there were entries like this for the CI tests) the unscheduled-maintenance-page will be used instead, while if there are no active appservers at all, the preliminary-page will be used.

**Testing instructions**

Grep the output of `consul kv get --recurse  ocim/instances | cut -f2- -d: | jq` on haproxy-integration.net.opencraft.hosting to find examples.

Check that those examples are properly rendered in `/etc/haproxy/haproxy.cfg` and that `/usr/sbin/haproxy -c -V -f /etc/haproxy/haproxy.cfg` is happy.

### DIFF
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -133,7 +133,7 @@ backend be-{{$config.domain_slug}}
     server unscheduled-maintenance-page 127.0.0.1:{{! maintenance_unscheduled_server_port !}}
       {{- end }}
     {{- else }}
-    server preliminary-page 149.202.180.183:80
+    server preliminary-page {{! PRELIMINARY_PAGE_SERVER_IP !}}:80
     http-request set-header Host 'default.opencraft.com'
     {{- end }}
     {{- end }}
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -126,8 +126,16 @@ backend be-{{$config.domain_slug}}
     {{- /* Loop over all active appservers for this instance. */ -}}
     {{- /* Each appserver in this list is a dict of the form {id: int, public_ip: str}. */ -}}
     {{- range $config.active_app_servers }}
+      {{- if .public_ip }}
     server appserver-{{ .id }} {{ .public_ip }}:80 cookie appserver-{{ .id }} {{ if $health_check }}check{{ end }}
+      {{- else}}
+    http-response redirect location / unless { capture.req.uri / }
+    server unscheduled-maintenance-page 127.0.0.1:{{! maintenance_unscheduled_server_port !}}
+      {{- end }}
+    {{- else }}
+    server preliminary-page 149.202.180.183:80
+    http-request set-header Host 'default.opencraft.com'
     {{- end }}
-    {{- end}}
+    {{- end }}
   {{- end }}
 {{- end }}


playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
```diff
@@ -126,8 +126,16 @@ backend be-{{$config.domain_slug}}
     {{- /* Loop over all active appservers for this instance. */ -}}
     {{- /* Each appserver in this list is a dict of the form {id: int, public_ip: str}. */ -}}
     {{- range $config.active_app_servers }}
+      {{- if .public_ip }}
     server appserver-{{ .id }} {{ .public_ip }}:80 cookie appserver-{{ .id }} {{ if $health_check }}check{{ end }}
+      {{- else}}
+    http-response redirect location / unless { capture.req.uri / }
+    server unscheduled-maintenance-page 127.0.0.1:{{! maintenance_unscheduled_server_port !}}
+      {{- end }}
+    {{- else }}
```
clemente: This is a `range` - `else`. I'm not familiar with Go's/consul-template's syntax and I didn't find this case in a quick skim of the documentation. I hope it works. Was it tested in practice?

SSPJ: Yep, I tested it. It works just the same as Python.

```
for x in []:
  # won't run
else:
  # will run
```

playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
```diff
@@ -126,8 +126,16 @@ backend be-{{$config.domain_slug}}
...
+    server preliminary-page 149.202.180.183:80
```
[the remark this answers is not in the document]
SSPJ: Updated.

SSPJ:
@clemente I addressed your comments. Good for another pass?

Note that this playbook fails on an unrelated step (upgrading nginx). I did not fix it manually because there is already a ticket to fix it in the playbook and it will be nice to have this server to test the fix on.

But it is a failure on master, not the code in this branch.

playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
clemente: That's not very reassuring :-) because the Python one does this:
```
for x in [1]:
  print("will run: for")
else:
  print("will run: else")
```
That is, both run. It's because the `else` runs after the iterations finish in a natural way (without break). There are more differences in error handling but this doesn't affect this template.

In this `range - else` in the template, I guess that in the end we output only 1 `server` directive (either the IP or the preliminary page), never 2. If that's tested, then it's fine.

clemente:
:+1: after checking the `range - else` (probably it's working fine since it did its job in practice).

  - [X] I tested this: I followed the instructions in the PR to check that the haproxy configurations are complete. I didn't redeploy this. I didn't do a full testing (see last comment and task)
  - [X] I read through the code, and discussed it
  - [X] I checked for accessibility issues
  - [] Includes documentation: no. But in this file we assume that the reader already knows what's an „unscheduled maintenance page“ (though from memory I don't know which one it is) and a „preliminary page“. There's documentation about „maintenance page“ earlier in the file.

playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
SSPJ: I got the info from [here](https://golang.org/pkg/text/template/):

> **{{range pipeline}} T1 {{else}} T0 {{end}}**
	The value of the pipeline must be an array, slice, map, or channel.
	If the value of the pipeline has length zero, dot is unaffected and
	T0 is executed; otherwise, dot is set to the successive elements
	of the array, slice, or map and T1 is executed.

And indeed, a look at `/etc/haproxy/haproxy.cfg` shows that all the backends have a single server line (except the backends with multiple appservers -- which have one per appserver).

PR #147, opened 2019-11-05
[SE-1791] Deploy the HA load balancers v2

SSPJ:
See [ansible-secrets !155](https://github.com/open-craft/ansible-secrets/pull/155).


playbooks/roles/load-balancer-v2/tasks/main.yml
```diff
@@ -25,51 +38,112 @@
...
       dest: "/lib/systemd/system/consul-certificates-render.service"
       mode: "0644"
+    - src: "cert-watcher.service"
+      dest: "/lib/systemd/system/cert-watcher.service"
+      mode: "0644"
 
-# Haproxy refuses to start when there are no valid SSL certificates in the
-# /etc/haproxy directory, so we copy a bogus cert there.
-- name: Copy snakeoil certificate
+- name: Copy over snakeoil certificate
```
giovannicimolin: Snakeoil certificates are present on Ubuntu under `/etc/ssl/certs/ssl-cert-snakeoil.pem` and `/etc/ssl/private/ssl-cert-snakeoil.key` if the `ssl-cert` package is present (I think this is a ubuntu default, so maybe we don't need to store it here). 
Do we need the same snakeoil certificate for all load balancers? Is is prossible to use the locally generated one?

SSPJ: We don't need the same one; it should never be used for anything except convincing HAProxy that it is okay to start. I have fixed this.

playbooks/roles/load-balancer-v2/defaults/main.yml
```diff
@@ -25,6 +25,9 @@ maintenance_provisioning_server_port: "{{ maintenance_server_port + 3 }}"
 maintenance_server_static_root: /var/www/maintenance
 maintenance_pages_repo: "https://github.com/open-craft/maintenance-pages"
 
+# The IP address below should always be IP address which the PRELIMINARY_PAGE_HOSTNAME resolves to.
+preliminary_page_server_ip: '149.202.180.183'
```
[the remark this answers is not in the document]
SSPJ: It's fine to leave public. Speaking from a security standpoint, treating any data that is trivially discoverable (such as DNS entries, port numbers, etc) as private is actually a bad idea: it gives the false security that our infrastructure is more private than it is.

Can we use the current host's IP? That's a good idea, I think. I will try that.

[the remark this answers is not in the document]
SSPJ: Or maybe not, on consideration. The comment says IP must match PRELIMINARY_PAGE_HOSTNAME. Not sure why, but I don't want to fiddle with it.

playbooks/roles/load-balancer-v2/files/prod/X29uZGVtYW5kLmNsb3VkZXJhLmNvbQ==.pem
```diff
@@ -0,0 +1,276 @@
+$ANSIBLE_VAULT;1.1;AES256
```
giovannicimolin: @SSPJ Just a quick note: I thought that these certificates were going to be stored on `ansible-common-secrets`.

playbooks/roles/load-balancer-v2/tasks/site-is-down.yml
```diff
@@ -0,0 +1,37 @@
+---
+# This set of tasks is meant to be run like so:
```
giovannicimolin: This is a nice playbook! :)
Can you update [the docs on this](https://gitlab.com/opencraft/documentation/private/blob/master/howtos/ocim/creating_maintenance_pages.md)?

playbooks/roles/load-balancer-v2/tasks/site-is-up.yml
```diff
@@ -0,0 +1,35 @@
+---
```
giovannicimolin: Same for this role, can you update the docs?

playbooks/roles/load-balancer-v2/files/prod/X29uZGVtYW5kLmNsb3VkZXJhLmNvbQ==.pem
SSPJ: @giovannicimolin No, the whole reason for introducing Ansible Vault was so that I could deploy these just like ordinary files.

playbooks/roles/load-balancer-v2/tasks/site-is-down.yml
SSPJ: Sure! I created https://gitlab.com/opencraft/documentation/private/merge_requests/422 .

As far as I know, those instructions are up-to-date, so I just made a note that you have to make the changes in the repository and linked to the deploy instructions on https://gitlab.com/opencraft/documentation/private/blob/master/ops/incidents.md#we-need-to-put-up-a-maintenance-page-for-some-site

playbooks/roles/load-balancer-v2/files/prod/X29uZGVtYW5kLmNsb3VkZXJhLmNvbQ==.pem
giovannicimolin: :+1:

SSPJ:
I have tested this, yep, and cleaned up the commit history.

PR #150, opened 2020-01-09
SE-2046 Allows client certificate to bypass SNI

pomegranited:
Adds symlink for `_<client site>.pem` to the Lets Encrypt-generated certificate for that site. Because of the leading underscore, this should be read before any other certificates, and treated as the default, thus bypassing SNI for this particular host.

This is needed because the client use Boomi Cloud for integration with their enrollment services, which [does not support SNI](https://community.boomi.com/s/feed/0D51W00006vwwoeSAA) and throws errors about certificates.

**Testing instructions**

Since this change just adds a file to the list of certificates already installed on haproxy prod servers, and since we don't really want to have to redeploy haproxy, these testing instructions are just to demonstrate that a new haproxy server deployment will Just Work™.

1. Ensure that the destination file matches the symlink that was added manually to the haproxy servers by checking the base64 decoded filename: 
   ```
   echo `echo "X29uZGVtYW5kLmNsb3VkZXJhLmNvbQo=" | base64 -D`.pem
   ```
1. Check that the contents of the new file matches the contents of the symlink file manually added to haproxy servers.
1. Check that the symlink installed on the haproxy servers points to the correct, valid client certificate file.

**Author Notes & Concerns**

* This trick will only work for one host, so if someone else needs us to bypass SNI, we'll have to come up with some other mechanism.
* Since this is a symlink, the file's contents unfortunately have to be in cleartext, and so will show the client's name.

**Reviewer**

- [ ] @giovannicimolin


pomegranited:
Closing this PR because the hack did not work.  Our default (first) certificate for the new haproxy servers is `/etc/haproxy/certs.pem`, a self-issued localhost cert.

PR #151, opened 2020-01-21
playbooks: replace deprecated with_items with loop

brett-oc:
Python 3.8.0 no longer parses lists in ansible roles properly: Users
were greeted with errors such as:

sre_constants.error: bad character range l-d at position 34

These mysterious errors were not present in the 3.7.x series and below.

Replacing the use of with_items: with loop: fixed the issue and the
playbook proceeded as expected.

Since use of with_* is no longer encouraged by Ansible [1], and the
unexpected behavior was exhibited by using with_items, replacement of
all instances of with_items seems the most reasonable. As with_items is
directly replacable with loop (unlike some of the other with_* items
like with_sequence), only the most ancient of Ansible versions could
run into trouble with the change.

[1] https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#migrating-from-with-x-to-loop

Fixes SE-2019

## Testing the changes

As there are no molecule tests for the roles that have been affected, the easiest way to test seems to be:

1. SSH into an Ubuntu 16.04 VM

2. Create/Activate a Python 3.8.x virtualenv

3. Install repo requirements (`pip install -r requirements.txt`)

4. Apply each role to verify no breakage.

### DIFF
--- a/playbooks/roles/pyenv/tasks/main.yml
+++ b/playbooks/roles/pyenv/tasks/main.yml
@@ -2,7 +2,7 @@
   apt:
     name: "{{ item }}"
     state: present
-  with_items: "{{ pyenv_prerequisite_packages }}"
+  loop: "{{ pyenv_prerequisite_packages }}"
 
 - name: Ensure that the "{{ pyenv_home }}" directory is present and has the right permissions
   file:


brett-oc:
This has been tested on Python 2.7.17, 3.59, and 3.8.1 with the following playbook featuring all the variants of looping that this repo uses:

```yaml
- name: Testing SE-2019
  hosts: localhost
  vars:
    BACKUP_PRUNER_TARSNAPPER_JOBS:
      - name: "mysql-backup"
        keypath: "/etc/keyfile"
        key: 'asdf'
        cache: "/var/cache/tarsnap"
        target: "asdf"
        deltas: 1h 1d 7d 30d 360d 18000d
    COMMON_SERVER_ETCKEEPER_COMMIT_USER: user
    COMMON_SERVER_ETCKEEPER_COMMIT_EMAIL: email
  tasks:
    - name: create dirs
      file:
        name: "{{ item }}"
        mode: 0700
        state: directory
      loop:
       - /etc/tarsnap
       - /var/cache

    - name: upload keys
      template:
        src: /tmp/a
        dest: "/etc/tarsnap/{{ item.name }}.key"
        mode: 0400
      loop: "{{ BACKUP_PRUNER_TARSNAPPER_JOBS }}"

    - name: Copy over cert-manager script and service files
      template:
        src: "{{ item.src }}"
        dest: "{{ item.dest }}"
        mode: "{{ item.mode }}"
      loop:
        - src: "/tmp/a"
          dest: "/usr/local/sbin/cert-manager.sh"
          mode: "0755"
        - src: "/tmp/a"
          dest: "/lib/systemd/system/cert-manager.service"
          mode: "0644"

    - name: etckeeper - configure git author settings
      git_config:
        name: "{{ item.name }}"
        value: "{{ item.value }}"
        scope: local
        repo: /etc
      loop:
        - name: user.name
          value: "{{ COMMON_SERVER_ETCKEEPER_COMMIT_USER }}"
        - name: user.email
          value: "{{ COMMON_SERVER_ETCKEEPER_COMMIT_EMAIL }}"

    - name: Create consul-template directories
      file:
        path: "{{ item.path }}"
        state: directory
        owner: "{{ item.owner }}"
        group: consul
        mode: 0750
      loop:
        - { path: "/etc/consul-template/config", owner: root }
        - { path: "/etc/consul-template/templates", owner: root }
```

lgp171188:
@brett-oc, my only concern with the changes in this PR is that it modifies a lot of files in multiple roles and testing this could be an issue. I would be more comfortable doing this only for the Ocim devstack setup code for the time-being and fix the compatibility issues with newer python versions when we decide to officially support them. CC @giovannicimolin as the devops specialist.

brett-oc:
@lgp171188, thanks for the feedback. Newbie opinion below:

Testing each of these roles seems not to be a trivial task, so unless I'm missing something I've tested as well as I can (as I don't have various accesses and don't know where servers are, etc). Fixing roles only as they come up would incur a non-trivial amount of debt: The user must first discover the bug, document the bug in a new ticket, and then have that ticket assigned/worked on piecemeal.

In my experiences replacing `with_items` with `loop` has always been a drop-in replacement - and the Ansible migration docs support that. I would argue the continuity risk to be minimal.

But I'm not intimately familiar with the stack. Take what you like of these humble words and leave the rest. :)

clemente:
@brett-oc, CC @giovannicimolin @lgp171188 
It looks [that](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#comparing-loop-and-with) `with_items` and `loop` have a different behaviour when an ansible `{{ VARIABLE }}` is a list of values.; I tested it in a local playbook, and indeed, `loop` grabs the whole item as 1 instead of flattening the list and processing each one. This can affect some of the cases in this PR, like for instance `loop: "{{ BACKUP_PRUNER_TARSNAPPER_JOBS }}"` (I see from your comment that you tried with 1 case, but will it work if there are many jobs in that variable?).
Adding `flatten` would help.

The documentation also mentions that `with_*` won't be deprecated, so we're not forced to migrate all of them.
It would be interesting to know why the original bug happened, because `with_items` should work fine if it's not deprecated.

So I'd go for a simpler version (i.e. fix just 1 case), otherwise it's hard to test each line changed by the PR.

pkulkark:
@brett-oc I agree with @lgp171188 and @clemente - I think it would be better to just fix the one case (i.e. ocim devstack) and maybe add a note in the readme (something like a troubleshooting section) so that anyone encountering the same error in other roles will know what to do.

giovannicimolin:
@brett-oc I agree with @lgp171188 and @clemente on the approach. Let's just fix this case and leave the other fixes for when we're working on those roles.

We have a lot of upgrades/fixes incoming so it shouldn't be long to fix most of these occurrences (and test them :smile:).

I've taken a note of this and will try to put on all devops tasks that change playbooks scheduled for the near future.

brett-oc:
I've updated the patch to only include the affected role in the ticket: pyenv

pkulkark:
Thanks @brett-oc. LGTM :+1:

- [x] I tested this: verified the error is no longer seen and devstack provisioning succeeds.
- [x] I read through the code
- [ ] I checked for accessibility issues: N/A
- [ ] Includes documentation: N/A
- [ ] I made sure any change in configuration variables is reflected in the corresponding
client's configuration-secure repository: N/A

PR #155, opened 2020-02-05
[BB-2052] Add a separate deploy hook script for deploying certificates to Consul

lgp171188:
This is needed because the provisioning is done by the certbot
installed inside the cert-manager's pipenv environment and the renewal
is handled by the systemd timer which uses the system-wide
installation of certbot. This script explicitly navigates to the
cert-manager directory and runs the hook script using 'pipenv run'.

**Testing instructions**:
* Deploy the changes in this PR on stage and verify that a well-formed `deploy_cert.sh` script is created under `/usr/local/sbin`.
* Deploy the corresponding changes in the `cert-manager` PR which set the deploy hook to the script added here.
* Restart the `cert-manager` service.
* Verify that for a new instance, the certificate is provisioned without any errors once the DNS record is created and the changes have propagated.
* Run a force-renewal for the certificate provisioned in the previous step from outside the `cert-manager` pipenv environment and verify that the certificate is renewed and the deploy hook is called without any issue.

### DIFF
--- a/playbooks/roles/cert-manager/tasks/main.yml
+++ b/playbooks/roles/cert-manager/tasks/main.yml
@@ -32,6 +32,12 @@
       dest: "/lib/systemd/system/cert-manager.service"
       mode: "0644"
 
+- name: Copy the deploy_cert.sh script template
+  template:
+    src: "deploy_cert.sh"
+    dest: "/usr/local/sbin/deploy_cert.sh"
+    mode: "0755"
+
 - name: Enable and start the cert-manager service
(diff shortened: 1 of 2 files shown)


lgp171188:
@giovannicimolin, I tested this on stage and it works. One issue that I noticed is that the certificate provisioning is attempted on stage as soon as an instance is created, even before the DNS records are set (which happens only during the provisioning of an appserver) and fails. A retry happens only when the appserver is activated and consul gets updated.

PR #156, opened 2020-02-12
[BB-2121] Redirect HTTP requests to HTTPS in Nginx configuration file

alanoe:
Work in Progress

- As I don't have access to the ansible-secrets repository, I'm not able to test the changes in the wordpress Ansible playbook. I have tested the solution in a separate Nginx webserver, installed from scratch, and it worked.

- TODO: create the corresponding SSL certificates and add them to ansible-secrets

### DIFF
--- a/playbooks/roles/wordpress/templates/default-nginx.j2
+++ b/playbooks/roles/wordpress/templates/default-nginx.j2
@@ -8,8 +8,9 @@ server {
 # HTTPS
 server {
         listen 443 ssl;
-        ssl_certificate /etc/ssl/certs/{{ DEFAULT_SITE_HOSTNAME }}.crt;
-        ssl_certificate_key /etc/ssl/private/{{ DEFAULT_SITE_HOSTNAME }}.key;
+        server_name {{ DEFAULT_SITE_HOSTNAME }};
+        ssl_certificate /etc/letsencrypt/live/{{ DEFAULT_SITE_HOSTNAME }}/fullchain.pem;
+        ssl_certificate_key /etc/letsencrypt/live/{{ DEFAULT_SITE_HOSTNAME }}/privkey.pem;
 
 	root /var/www/{{ DEFAULT_SITE_HOSTNAME }};
 	index index.html index.php;
--- a/playbooks/roles/wordpress/templates/default-nginx.j2
+++ b/playbooks/roles/wordpress/templates/default-nginx.j2
@@ -1,6 +1,15 @@
+# HTTP
 server {
 	listen *:80;
 	server_name {{ DEFAULT_SITE_HOSTNAME }};
+	return 301 https://$host$request_uri;
+}
+
+# HTTPS
+server {
+        listen 443 ssl;
+        ssl_certificate /etc/ssl/certs/{{ DEFAULT_SITE_HOSTNAME }}.crt;
+        ssl_certificate_key /etc/ssl/private/{{ DEFAULT_SITE_HOSTNAME }}.key;
 
 	root /var/www/{{ DEFAULT_SITE_HOSTNAME }};
 	index index.html index.php;
--- a/playbooks/roles/wordpress/templates/maintenance-nginx.j2
+++ b/playbooks/roles/wordpress/templates/maintenance-nginx.j2
@@ -1,6 +1,15 @@
+# HTTP
 server {
 	listen *:80;
         server_name {{ MAINTENANCE_SITE_HOSTNAME }};
+        return 301 https://$host$request_uri;
+}
+
+# HTTPS
+server{
+        listen 443 ssl;
+        ssl_certificate /etc/ssl/certs/{{ MAINTENANCE_SITE_HOSTNAME }}.crt;
+        ssl_certificate_key /etc/ssl/private/{{ MAINTENANCE_SITE_HOSTNAME }}.key;
 
 	root /var/www/{{ MAINTENANCE_SITE_HOSTNAME }};
 	index index.html index.php;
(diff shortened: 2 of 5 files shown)


playbooks/roles/wordpress/templates/default-nginx.j2
```diff
@@ -1,6 +1,15 @@
+# HTTP
 server {
 	listen *:80;
 	server_name {{ DEFAULT_SITE_HOSTNAME }};
+	return 301 https://$host$request_uri;
+}
+
+# HTTPS
+server {
+        listen 443 ssl;
```
giovannicimolin: Missed server name:
```suggestion
        listen *:443 ssl;
        server_name {{ DEFAULT_SITE_HOSTNAME }};
```

playbooks/roles/wordpress/templates/default-nginx.j2
```diff
@@ -1,6 +1,15 @@
...
+        ssl_certificate /etc/ssl/certs/{{ DEFAULT_SITE_HOSTNAME }}.crt;
```
giovannicimolin: The certificates provisioned by certbot reside in `/etc/letsencrypt/live/DOMAIN_NAME/`, so you should change this to:
```suggestion
        ssl on;
        ssl_certificate /etc/letsencrypt/live/{{ DEFAULT_SITE_HOSTNAME }}/fullchain.pem;
        ssl_certificate_key /etc/letsencrypt/live/{{ DEFAULT_SITE_HOSTNAME }}/privkey.pem;
```

giovannicimolin: Don't forget to remove the line below if accepting this suggestion.

playbooks/roles/wordpress/templates/maintenance-nginx.j2
```diff
@@ -1,6 +1,15 @@
+# HTTP
```
giovannicimolin: Same as the comments on the other role.

playbooks/roles/wordpress/templates/default-nginx.j2
alanoe: +1 to the certificates path. 

Regarding the "ssl on" directive, the ssl directive was made obsolete in Nginx 1.15.0 and Nginx doc recommends to use the ssl parameter of the listen directive  instead. Reference: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl

playbooks/roles/mount-web-directory/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
```
giovannicimolin: I think we don't need a separate role. You can use ansible tags to separate tasks.
Move this task to inside the wordpress role and set a tag like `only-mount`, then run the common server, and then run the wordpress role with the remaining tasks again (maybe using another tag).
Example:
```
---
- name: Deploy the OpenCraft landing page
  hosts: wordpress
  become: true
  vars:
        ....
  roles:
    - role: wordpress
      tags: only-mount

    - common-server
    - geerlingguy.php-versions
    - geerlingguy.php-mysql

    - role: wordpress
      tags: install-deploy

    - tarsnap
```

giovannicimolin: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html

playbooks/wordpress.yml
```diff
@@ -23,6 +23,7 @@
       - php7.3-json
     - php_mysql_package: php7.3-mysql
   roles:
+    - role: mount-web-directory
     - role: common-server
       tags: 'common-server'
       vars:
```
giovannicimolin: Certbot can run as usual here, you can remove these lines:
```
# certbot must run after deploying the wordpress role
COMMON_SERVER_INSTALL_CERTBOT: false
```

giovannicimolin: You should also remove the `- certbot` line below.

playbooks/roles/mount-web-directory/tasks/main.yml
alanoe: Indeed both solutions work. IMHO smaller roles with well-defined scope are easier to read and maintain (for instance, the "wordpress" role should only install the Wordpress content management system and not execute machine-specific tasks, e.g. mount the /var/www directory). Is there any benefit of using tags here besides reducing the number of roles?

giovannicimolin: These playbooks are old and carry a lot of drift with them.

My main concern by splitting the role is that we'd need to add more configuration and secret variables for a proper implementation, passing `/dev/sdX', mount point and permissions of the attached volume from the `ansible-secrets` repo instead of hardcoding them in the role.

I can accept the work as is after you add a README with some context on the role, but it's not the ideal solution.

alanoe: Reading your last comment, I have remembered that the purpose of the Ansible role is to be reusable, so it makes sense. I have changed my patch to use "pre_tasks" instead of a separate role.

playbooks/wordpress.yml
```diff
@@ -22,15 +22,21 @@
       - php7.3-imap
       - php7.3-json
     - php_mysql_package: php7.3-mysql
+  pre_tasks:
```
giovannicimolin: Nice work using the pre-tasks here. :+1: 
We usually leave all tasks to be run inside the roles, but this is an acceptable one off (we only mount the `www` dir while deploying wordpress.

PR #162, opened 2020-03-31
[ARCHIVE] SE-2197: add an `apticron` role

tartansandal:
**Keep this PR in case we need it at some time in the future, even though we don't want to merge it right now.**

Adds a role to install and configure apticron role.  

This role takes an optional 'email_to' parameter which defaults to 'ops@opencraft.com'.

Considering using `inventory_hostname` rather than relying on `hostname -f` for the host name identification in the report. Please let me know if you think this is a good idea.

**Jira tickets:** SE-2197

**Testing instructions:**

You can test this from your OCIM vagrant devstack.
I've added the role to the `ocim-vagrant.yml` playbook for testing/review, so you can 
```
ANSIBLE_INVENTORY=.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory
ansible-playbook deploy/playbooks/ocim-vagrant.yml -t apticron
```
We can role that back before merging and perhaps add the role to `ocim` and `cert-manger` playbooks.

**Reviewers**
- [ ] @swalladge 
- [ ] @lgp171188

### DIFF
--- a/playbooks/roles/apticron/templates/apticron.conf.j2
+++ b/playbooks/roles/apticron/templates/apticron.conf.j2
@@ -36,7 +36,8 @@ EMAIL="{{ email_to }}"
 # of "hostname -f" for the system name in the mails it generates. This option
 # overrides the ALL_FQDNS above.
 #
-# XXX Consider using the ansible name.
+# XXX If you are having trouble with the system names returned by `hostname -f`,
+# consider using the hostname used by the ansible inventory.
 # SYSTEM="{{ inventory_hostname }}"
 
 #
--- a/playbooks/roles/apticron/templates/apticron.conf.j2
+++ b/playbooks/roles/apticron/templates/apticron.conf.j2
@@ -99,4 +99,4 @@ EMAIL="{{ email_to }}"
 # 'From:' field used in the notification e-mails. Your default sender will
 # be something like root@ubuntu1604.localdomain.
 #
-# CUSTOM_FROM=""
+CUSTOM_FROM="{{ email_from }}"
--- a/playbooks/roles/apticron/templates/apticron.conf.j2
+++ b/playbooks/roles/apticron/templates/apticron.conf.j2
@@ -59,7 +59,7 @@ EMAIL="{{ email_to }}"
 # packages on hold in your system. The default behavior is downloading and
 # listing them as any other package.
 #
-# NOTIFY_HOLDS="0"
+NOTIFY_HOLDS="0"
 
 #
 # Set NOTIFY_NEW="0" if you don't want to be notified about packages which
@@ -70,13 +70,13 @@ EMAIL="{{ email_to }}"
 # so they will be listed in dist-upgrade output. Please take a look at
 # http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=531002#44
 #
-# NOTIFY_NEW="0"
+NOTIFY_NEW="0"
 
 #
 # Set NOTIFY_NO_UPDATES="0" if you don't want to be notified when there is no
 # new versions. Set to 1 could assure you that apticron works well.
 #
-# NOTIFY_NO_UPDATES="0"
+NOTIFY_NO_UPDATES="0"
 
 #
 # Set CUSTOM_SUBJECT if you want to replace the default subject used in
--- a/playbooks/roles/apticron/templates/apticron.conf.j2
+++ b/playbooks/roles/apticron/templates/apticron.conf.j2
@@ -1,0 +1,101 @@
+# apticron.conf
+#
+# set EMAIL to a space separated list of addresses which will be notified of
+# impending updates
+#
+EMAIL="{{ email_to }}"
+
+#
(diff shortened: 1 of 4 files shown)


playbooks/roles/apticron/templates/apticron.conf.j2
```diff
@@ -0,0 +1,101 @@
...
+#
+# ALL_FQDNS="1"
+
+#
+# Set SYSTEM if you would like apticron to use something other than the output
+# of "hostname -f" for the system name in the mails it generates. This option
+# overrides the ALL_FQDNS above.
+#
+# XXX Consider using the ansible name.
+# SYSTEM="{{ inventory_hostname }}"
```
swalladge, line 40: @tartansandal This seems fine. I don't think it makes a huge amount of difference here? It probably could be even easier to leave it out and let apticron use `hostname -f`.

Note also "This option overrides the ALL_FQDNS above."

playbooks/roles/apticron/templates/apticron.conf.j2
```diff
@@ -0,0 +1,101 @@
...
+# $CUSTOM_NO_UPDATES_SUBJECT='[apticron] $SYSTEM: no updates'
+#
+# CUSTOM_NO_UPDATES_SUBJECT=""
+
+#
+# Set CUSTOM_FROM if you want to replace the default sender by changing the
+# 'From:' field used in the notification e-mails. Your default sender will
+# be something like root@ubuntu1604.localdomain.
+#
+# CUSTOM_FROM=""
```
swalladge, line 101: I'm not sure if we would need this or not? Not sure what our mail sending setup is on various machines. CC @lgp171188

playbooks/roles/apticron/templates/apticron.conf.j2
tartansandal: All but the very first option (EMAIL) is commented out. The commented options are not the defaults, but suggestions. So `# ALL_FQDNS=1` is just a suggestion if you don't want the default of using `hostname -f`.  So the override is only relevant if that is commented out.

The `# SYSTEM` suggestion is just me being helpful to a future developer mucking around with this role. :smile: I'll make the comment more explicit, say, 
```
# XXX If you are having trouble with the system names returned by `hostname -f`,
# consider using the hostname used by the ansible inventory.
# SYSTEM="{{ inventory_hostname }}"
```

playbooks/roles/apticron/templates/apticron.conf.j2
tartansandal: All of this is just comments from the default configuration file. I like to leave those in so future developers hacking the ansible role don't have to dig through to much documentation.

tartansandal:
> @tartansandal I ran this against my ocim devstack, and it installed perfectly. However, I don't see any cron jobs set up. I would have expected to see `/etc/cron.daily/apticron` present. Does that need to be added manually?

That gets set up by the package under `/etc/cron.d/apitcron`. The contents are:
```
# cron entry for apticron

35 * * * * root if test -x /usr/sbin/apticron; then /usr/sbin/apticron --cron; else true; fi
```
At a guess, the 35 is probably randomly generated during install to guard against the "thundering heard" problem, though I have not checked that.

playbooks/roles/apticron/templates/apticron.conf.j2
swalladge: Ah I missed that they were commented. :man_facepalming:  All good! :)

playbooks/roles/apticron/templates/apticron.conf.j2
swalladge: Yep, but I was meaning that we may actually need to set this depending on our mail setup on the server we want to deploy this to.

swalladge:
@tartansandal 

>   That gets set up by the package under /etc/cron.d/apitcron. 

Ah nice. Can confirm it was set up for me too. :+1: It was also a different minute number (40) for me, so you're probably right about it being random.

playbooks/roles/apticron/templates/apticron.conf.j2
tartansandal: Very good point @swalladge :+1:  I just added a patch so the `CUSTOM_FROM` can be set via role parameters (defaulting to something sensible). I've added some dummy settings to the `ocim-vagrant.yml` playbook so we can test that this works.

tartansandal:
@swalladge I've parameterised CUSTOM_FROM and tweaked the configuration file to reduce the potential number of emails sent as per the discussion on https://tasks.opencraft.com/browse/SE-2197.
I know we are not going to use this role for this task, but I think it still maybe useful to have around.
Am I good to rebase (removing the ocim-vagrant hacks) and merge?

tartansandal:
@lgp171188 @swalladge 

> ... is not good to merge code that we know we won't use and recommended keeping this PR open so that if it is needed in the future, the changes are readily available.

All good :+1: I'll adjust the PR name and add a comment so we don delete this accidentally :wink:

PR #167, opened 2020-04-16
[BB-2278] Implement changes to allow consul auto configuration on AWS

lgp171188:
This PR implements the changes to allow automatic generation of consul configuration at service startup time when the `consul_auto_generate_config` flag is set to a truthy value.

**Testing instructions**:
* Spawn an EC2 instance and run the sidecar playbook with the `consul_auto_generate_config` flag set to `true` in the `hosts` file.
* Verify that the playbook finishes successfully. Log in to the EC2 instance and verify that the auto-generated consul configuration file, `/etc/consul/config.json` uses the correct public IP address as the value of the `advertise_addr` parameter and that the `node_name` parameter is set to the format `<consul_nodename>-<private IP address separated by - instead .>-<public IP address separated by - . instead of .>`.
- Verify that the systemd service unit file `/etc/systemd/system/consul.service` uses the wrapper script `/usr/local/sbin/run_consul.sh` instead of the default `https://github.com/open-craft/ansible-playbooks/pull/new/guruprasad/BB-2278-autodetect-consul-configuration`.
- Verify that the `run_script.sh` script and `update_consul_config.py` scripts are copied to `/usr/local/sbin` directory.
* Check the logs of the consul service for any errors by running `sudo journalctl -f -u consul.service`.
* Run `consul members` on the instance to verify that it is added to the cluster. Verify the same from another node in the same cluster.
* Stop the EC2 instance. Verify from another node in the same consul cluster that the shutdown instance is now showing up with status as `left`.
* Start the EC2 instance after a few minutes.
* Check if its private/public IP addresses have changed.
* If either address has changed, verify that the consul configuration file `/etc/consul/config.json` has all its parameters updated to match the updated IP addresses.
* Repeat the steps mentioned before to verify the consul membership and the status of the instance in the cluster.
* Optionally, attach an elastic IP address to the instance and restart the consul service. Verify that the configuration has been updated properly, the service is running properly and the consul cluster membership has been updated appropriately by following the steps mentioned before.
* Create an AMI from the EC2 instance and create a new auto-scaling group. Set the created AMI in the launch template.
* Launch a new EC2 instance via the auto-scaling group. Once it is ready, check and verify its consul configuration, service status and cluster membership using the steps mentioned before.
* Also verify that the existing behaviour is retained when `consul_auto_generate_config` is not specified or set to a falsey value.

* This PR introduces yet another flag and adds to the existing clutter.
* A service restart is required to detect IP address changes on a running EC2 instance, for example, when an elastic IP address is attached or removed.
* Works only on AWS.

### DIFF
--- a/playbooks/roles/consul/README.md
+++ b/playbooks/roles/consul/README.md
@@ -1,27 +1,31 @@
 # Ansible Role for Consul
 
-Description N/A
-
-## Requirements
-
-N/A
+This role installs and sets up Consul to connect to a Consul cluster.
 
 ## Role Variables
 
-N/A
+See `defaults/main.yml`.
 
 ## Dependencies
 
-None.
+This depends on the 'nginx-proxy' role in the same repository.
 
-## Example Playbook
+## Usage
 
-Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+This role usually requires the `consul_ip`, `consul_nodename` variables to be set for each host. However, in AWS,
+when EC2 instances are launched from an AMI or stopped and started, or use an elastic IP, their IP addresses will
+likely change. So to allow automatic configuration of the `consul_ip`, `consul_nodename` variables
+while generating the consul configuration, the `consul_auto_generate_config` flag can be set to a
+truthy value.
 
-    - hosts: servers
-      roles:
-        - consul
+When `consul_auto_generate_config` is set to a truthy value, a script which wraps the consul startup command is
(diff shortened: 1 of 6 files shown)


lgp171188:
@giovannicimolin, I have added documentation in the `README` file of the role. Please check and let me know if it is okay. If you don't have the time to review this before the sprint ends, I will merge the PR. Any documentation issues can be addressed in a follow-up task.

PR #168, opened 2020-04-17
Update python version from 3.6.9 to 3.6.10

pkulkark:
This PR bumps the python version from 3.6.9 to 3.6.10. This also requires an update to pyenv to the [version](https://github.com/pyenv/pyenv/blob/master/CHANGELOG.md#1216) that supports python 3.6.10.


lgp171188:
@pkulkark, the changes in your PR look good. I have also added a fix for a minor issue that I noticed when provisioning a new devstack (which was also noticed and reported by the recent newcomers).

PR #173, opened 2020-06-18
[BB-2527] Add a role for deploying a Matomo instance

lgp171188:
**Testing instructions**:
* Deploy a Matomo instance using this role or alternatively test an already deployed instance (see the corresponding `ansible-secrets` PR)


playbooks/roles/matomo/templates/matomo_archive_reports.sh.j2
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker exec -u www-data matomo php /var/www/html/console core:archive --url=https://{{ MATOMO_APP_DOMAIN}} > /dev/null
```
viadanna, line 3: Can you please add some documentation to this script, just a point on what it does and why it's necessary.

It's easy to forget all about this when we stop touching this instance.

lgp171188: @viadanna, done. :+1:

PR #175, opened 2020-07-15
[SE-2948] Remove redirect to preliminary page

mavidser:
When no appservers are active, instance LMS and Studio URLs should not redirect the user, but just display contents of default.opencraft.com. Have applied the patch manually to the below instance on stage too, for testing purposes.

**JIRA tickets**: [SE-2948](https://tasks.opencraft.com/browse/SE-2948)

**Sandbox URL**: https://sid-test-2.stage.opencraft.hosting (The DNS has 3 IP addresses, so you might see the old deployment on IPs other than 54.36.101.120)

**Testing instructions**:

1. Run `curl --insecure -v --header "Host:se-2870-test.opencraft.hosting" https://54.36.101.120` and verify 301 Moved Permanently response.
2. Run `curl --insecure -v --header "Host:sid-test-2.stage.opencraft.hosting" https://54.36.101.120` and verify 200 OK is returned, with content of default.opencraft.com
3. Verify that `backend be-edxins-sid-test-2sta-567` section of `/etc/haproxy/haproxy.cfg` on stage haproxy has the updated code.

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -135,7 +135,7 @@ backend be-{{$config.domain_slug}}
     server unscheduled-maintenance-page 127.0.0.1:{{! maintenance_unscheduled_server_port !}}
       {{- end }}
     {{- else }}
-    server preliminary-page {{! preliminary_page_server_ip !}}:80
+    server preliminary-page {{! preliminary_page_server_ip !}}:443 ssl verify required ca-file /etc/ssl/certs/ca-certificates.crt
     http-request set-header Host 'default.opencraft.com'
     {{- end }}
     {{- end }}


mtyaka:
Thanks @mavidser! I'm currently getting a 503 service unavailable for https://sid-test-2.stage.opencraft.hosting/

mavidser:
@mtyaka I manually applied changes to that instance's haproxy config, but it looks like they get overwritten whenever someone deploys something new on stage. I have applied those changes again, but depending on when you test it, they might get reverted again. If the curl fails, you'd have to change `/etc/haproxy/haproxy.cfg` on stage haproxy again to test it out. 

Keep in mind that the DNS might use different IPs and I've only updated the config on `haproxy-stage-1 (54.36.101.120)`, so you'd have to do a couple of hard refreshes in browser, or use the mentioned curl command above.

_Diff of my changes on `/etc/haproxy/haproxy.cfg` right now_:
```diff
312c312,313
<     server appserver-2035 147.135.193.185:80 cookie appserver-2035
---
>     server preliminary-page 149.202.180.183:443 ssl verify required ca-file /etc/ssl/certs/ca-certificates.crt
>     http-request set-header Host 'default.opencraft.com'
```

PR #176, opened 2020-07-22
[BB-2720] Add playbook and role for deploying WordPress using docker

lgp171188:
This PR adds a playbook and a role to deploy a WordPress instance using docker. The nginx reverse proxy configuration used in this role is based on the configuration at https://www.nginx.com/resources/wiki/start/topics/recipes/wordpress/.


playbooks/wordpress-docker.yml
```diff
@@ -0,0 +1,24 @@
...
+- name: Deploy a WordPress instance using docker
+  hosts: wordpress-docker
+  become: true
+  pre_tasks:
+      # Required by the consul role.
+    - name: Ensure that unzip is installed
+      apt:
+        name: unzip
+        state: present
+        update_cache: yes
```
ahmedbilal, line 11: Shouldn't it live in `consul` role than?

lgp171188: @ahmedbilal, this could live in the `consul` role, yes. But I was trying to avoid making changes to any existing roles. It works okay on all other playbooks because `unzip` is automatically installed by the `common-server` role that is included in all the other playbooks.

I will update the PR to add this to the `consul` role itself since it is an explicit dependency.

PR #177, opened 2020-08-12
[BB-2851] Huey queue monitor

shimulch:
#### Description
Adds a new ``redis-exporter`` role to send ``redis`` metrics to ``prometheus``. It uses [prometheus_redis_exporter_role](https://github.com/idealista/prometheus_redis_exporter_role) as dependency and adds nginx & consul configurations to work with Ocim infrastructure.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "playbooks/roles/geerlingguy.php-versions"]
 	path = playbooks/roles/geerlingguy.php-versions
 	url = https://github.com/geerlingguy/ansible-role-php-versions
+[submodule "playbooks/roles/prometheus_redis_exporter_role"]
+	path = playbooks/roles/prometheus_redis_exporter_role
+	url = https://github.com/idealista/prometheus_redis_exporter_role
--- a/playbooks/group_vars/prometheus/public.yml
+++ b/playbooks/group_vars/prometheus/public.yml
@@ -35,6 +35,9 @@ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD: 80
 PROMETHEUS_ALERT_HTTP_RESPONSE_TIME_THRESHOLD: 1
 PROMETHEUS_ALERT_HTTP_RESPONSE_FRONTEND_ERROR_THRESHOLD: 70
 
+OCIM_QUEUE_REDIS_KEY: null
+PROMETHEUS_ALERT_OCIM_QUEUE_SIZE_THRESHOLD: null
+
 prometheus_rules:
   - name: NodeRules
     rules:
@@ -161,7 +164,17 @@ prometheus_rules:
       annotations:
         summary: "MongoDB instance [[ $labels.node ]] ([[ $labels.instance ]]) down."
         description: "[[ $labels.node ]] ([[ $labels.instance ]]) of job [[ $labels.job ]] is down."
-
+  - name: "OcimQueueSizeRule"
+    rules:
+    - alert: "ocim_queue_too_big"
+      expr: 'redis_key_size{key="{{ OCIM_QUEUE_REDIS_KEY }}"} > {{ PROMETHEUS_ALERT_OCIM_QUEUE_SIZE_THRESHOLD }}'
+      for: "0s"
+      labels:
+        severity: "critical"
+        environment: "[[ $labels.environment ]]"
+      annotations:
+        summary: "Ocim queue overloaded."
+        description: "Ocim queue overloaded."
 # ALERT MANAGER ################################################################
 
 prometheus_alert_manager_web__listen_address: "127.0.0.1:9093"
--- a/playbooks/ocim.yml
+++ b/playbooks/ocim.yml
@@ -9,3 +9,9 @@
 
     - role: ocim
       tags: 'ocim'
+
+    - role: redis-exporter
+      tags: 'redis-exporter'
+      REDIS_EXPORTER_CHECK_KEYS: "{{ OCIM_QUEUE_REDIS_KEY }}"
+      REDIS_EXPORTER_REDIS_ADDR: "{{ OCIM_REDIS_ADDRESS }}"
+      when: REDIS_EXPORTER_PASSWORD is not none
--- a/playbooks/roles/redis-exporter/README.md
+++ b/playbooks/roles/redis-exporter/README.md
@@ -1,0 +1,18 @@
+redis-exporter
+==============
+
+Extends [prometheus_redis_exporter_role](https://github.com/idealista/prometheus_redis_exporter_role)
+to expose the size of the specified Redis key.
+
+Role Variables
+--------------
+
+Check available variables here:
+- `playbooks/roles/prometheus_redis_exporter_role/defaults/main.yml`
+- `playbooks/roles/redis-exporter/defaults/main.yml`
+
+Dependencies
+------------
+
+- nginx-proxy
+- [prometheus_redis_exporter_role](https://github.com/idealista/prometheus_redis_exporter_role)
--- a/playbooks/roles/redis-exporter/defaults/main.yml
+++ b/playbooks/roles/redis-exporter/defaults/main.yml
@@ -1,0 +1,20 @@
+---
+# The basic auth password to access the redis exporter.
+REDIS_EXPORTER_PASSWORD: null
+
+REDIS_EXPORTER_NGINX_CONFIG_PATH: /etc/nginx/sites-available/redis-exporter
+REDIS_EXPORTER_HTPASSWD_PATH: /etc/nginx/htpasswd
+
+REDIS_EXPORTER_USE_SSL: true
+
+REDIS_EXPORTER_REDIS_ADDR: null
+REDIS_EXPORTER_CHECK_KEYS: null
+
+REDIS_EXPORTER_CONSUL_SERVICE:
+  service:
+    name: redis-exporter
+    tags:
+      - redis-exporter
+      - "{{ 'monitor-https' if REDIS_EXPORTER_USE_SSL else 'monitor' }}"
+    port: 19103
+    enable_tag_override: true
--- a/playbooks/roles/redis-exporter/meta/main.yml
+++ b/playbooks/roles/redis-exporter/meta/main.yml
@@ -1,0 +1,17 @@
+---
+  dependencies:
+    - name: prometheus_redis_exporter_role
+      redis_exporter_redis_addr: "{{ REDIS_EXPORTER_REDIS_ADDR }}"
+      redis_exporter_check_keys: "{{ REDIS_EXPORTER_CHECK_KEYS }}"
+    - name: nginx-proxy
+      NGINX_PROXY_HTPASSWD_PATH: "{{ REDIS_EXPORTER_HTPASSWD_PATH }}"
+      # The username is a bit misleading – should actually be "prometheus", since it's the server
+      # authenticating, and the same username is used for scraping other exporters, but it's not worth
+      # the hassle to change it.
+      NGINX_PROXY_USERNAME: node-exporter
+      NGINX_PROXY_PASSWORD: "{{ REDIS_EXPORTER_PASSWORD }}"
+      NGINX_PROXY_AUTH_DOMAIN: Prometheus Redis Exporter
+      NGINX_PROXY_CONFIG_PATH: "{{ REDIS_EXPORTER_NGINX_CONFIG_PATH }}"
+      NGINX_PROXY_PUBLIC_PORT: 19103
+      NGINX_PROXY_INTERNAL_PORT: 9121
+      NGINX_PROXY_USE_SSL: "{{ REDIS_EXPORTER_USE_SSL }}"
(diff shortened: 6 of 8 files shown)


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "playbooks/roles/geerlingguy.php-versions"]
 	path = playbooks/roles/geerlingguy.php-versions
 	url = https://github.com/geerlingguy/ansible-role-php-versions
+[submodule "playbooks/roles/prometheus_redis_exporter_role"]
```
lgp171188, line 34: @shimulch, this role installs https://github.com/oliver006/redis_exporter which is a lot more generic than what is needed for this task. So it might be better to use something more specific to monitoring the queues like https://github.com/ashishtiwari1993/redis_queue_length_exporter. Please check and pick a maintained alternative CC @0x29a

shimulch: @lgp171188 I've been using [REDIS_EXPORTER_CHECK_KEYS](https://github.com/open-craft/ansible-playbooks/pull/177/commits/9ee8a704467cefbe0cfe42e05f79a21b7e6a0112#diff-7b58e4e987cec3b7a4d4eeb3c36094d4R11) to export length of a specific key, which is documented [here](https://github.com/oliver006/redis_exporter#command-line-flags). It should do the same task as ``llkey`` implementation of ``redis_queue_length_exporter`` (need to test).  

Also [redis_exporter](https://github.com/oliver006/redis_exporter) is better maintained than [redis_queue_length_exporter](https://github.com/ashishtiwari1993/redis_queue_length_exporter) and documented in [prometheus doc](https://prometheus.io/docs/instrumenting/exporters/#databases).

I don't see any other exporter that only exports lengths other than your shared one. Do you want me to change it?

playbooks/group_vars/prometheus/public.yml
```diff
@@ -35,6 +35,8 @@ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD: 80
 PROMETHEUS_ALERT_HTTP_RESPONSE_TIME_THRESHOLD: 1
 PROMETHEUS_ALERT_HTTP_RESPONSE_FRONTEND_ERROR_THRESHOLD: 70
 
+PROMETHEUS_ALERT_HUEY_QUEUE_SIZE_THRESHOLD: 200
```
lgp171188, line 38: @shimulch, can you name this variable so that it is obvious that this is related to Ocim's queue?

shimulch: @lgp171188 ``PROMETHEUS_ALERT_OCIM_HUEY_QUEUE_SIZE_THRESHOLD`` seems a better name?

0x29a: @shimulch I think `PROMETHEUS_ALERT_OCIM_HUEY_QUEUE_SIZE_THRESHOLD` should be good. Can you please add change with rename?

shimulch: @0x29a updated.

playbooks/group_vars/prometheus/public.yml
```diff
@@ -35,6 +35,8 @@ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD: 80
 PROMETHEUS_ALERT_HTTP_RESPONSE_TIME_THRESHOLD: 1
 PROMETHEUS_ALERT_HTTP_RESPONSE_FRONTEND_ERROR_THRESHOLD: 70
 
+PROMETHEUS_ALERT_OCIM_HUEY_QUEUE_SIZE_THRESHOLD: 200
```
lgp171188, line 38: @shimulch, @0x29a, since this exporter is used only on one specific server, can we move this to the specific file under the `host_vars` directory of the `ansible-secrets` repository? Since Shimul doesn't have access to that repo, you will have to do that Demid.

playbooks/group_vars/prometheus/public.yml
```diff
@@ -161,7 +163,17 @@ prometheus_rules:
       annotations:
         summary: "MongoDB instance [[ $labels.node ]] ([[ $labels.instance ]]) down."
         description: "[[ $labels.node ]] ([[ $labels.instance ]]) of job [[ $labels.job ]] is down."
-
+  - name: "HueyRules"
```
lgp171188, line 166: @shimulch, @0x29a there should be nothing specific to Huey (Rename all instances Huey to say Redis queue or Redis list) or Ocim here and the queue key name should not be hardcoded and instead be specified using a variable.

playbooks/roles/redis-exporter/defaults/main.yml
```diff
@@ -0,0 +1,20 @@
...
+# The basic auth password to access the redis exporter.
+REDIS_EXPORTER_PASSWORD: null
+
+REDIS_EXPORTER_NGINX_CONFIG_PATH: /etc/nginx/sites-available/redis-exporter
+REDIS_EXPORTER_HTPASSWD_PATH: /etc/nginx/htpasswd
+
+REDIS_EXPORTER_USE_SSL: true
+
+REDIS_EXPORTER_REDIS_ADDR: 'redis://localhost:6379'
+REDIS_EXPORTER_CHECK_KEYS: 'huey.redis.opencraft'
```
lgp171188, line 11: @shimulch, @0x29a, don't hard-code this here. It should be configured using Ansible host_vars or group_vars that apply only to Ocim servers.

playbooks/roles/redis-exporter/meta/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    - name: nginx-proxy
+      NGINX_PROXY_HTPASSWD_PATH: "{{ REDIS_EXPORTER_HTPASSWD_PATH }}"
+      # The username is a bit misleading – should actually be "prometheus", since it's the server
+      # authenticating, and the same username is used for scraping other exporters, but it's not worth
+      # the hassle to change it.
+      NGINX_PROXY_USERNAME: node-exporter
+      NGINX_PROXY_PASSWORD: "{{ REDIS_EXPORTER_PASSWORD }}"
+      NGINX_PROXY_AUTH_DOMAIN: Prometheus Redis Exporter
+      NGINX_PROXY_CONFIG_PATH: "{{ REDIS_EXPORTER_NGINX_CONFIG_PATH }}"
+      NGINX_PROXY_PUBLIC_PORT: 19121
```
lgp171188, line 15: @shimulch, @0x29a, this port should be opened in the firewall using `ufw`. To avoid adding new ports for each service, is it possible to use virtual hosts listening on the same port to serve multiple sites?

playbooks/group_vars/prometheus/public.yml
giovannicimolin: We can either go with `redis` queued tasks or a very specific `Ocim queue size` alert so that the FFs/ops reviewer don't need extra context to know what's going on.

shimulch: Changed to ``OcimQueueSizeRule``. I guess now ``OCIM_QUEUE_REDIS_KEY`` & ``PROMETHEUS_ALERT_OCIM_QUEUE_SIZE_THRESHOLD`` have to move to ``secrets`` repository.

playbooks/roles/redis-exporter/defaults/main.yml
shimulch: @lgp171188 Removed. Now those are being passed as variable from [Playbook](https://github.com/open-craft/ansible-playbooks/pull/177/commits/116bfa7a889210bb633d092ee16aab2fb878858d#diff-ec9878bb813db60641651149cd4209d3). I think those variables will be set on ``secrets`` repository?

playbooks/roles/redis-exporter/meta/main.yml
shimulch: @lgp171188 I think we can make use of ``NGINX_PROXY_SERVER_NAME`` config from ``nginx-proxy`` role. But the server name might need to be passed from ``secrets`` repository, just like Ocim domain names get passed down.

playbooks/group_vars/prometheus/public.yml
```diff
@@ -35,6 +35,9 @@ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD: 80
 PROMETHEUS_ALERT_HTTP_RESPONSE_TIME_THRESHOLD: 1
 PROMETHEUS_ALERT_HTTP_RESPONSE_FRONTEND_ERROR_THRESHOLD: 70
 
+OCIM_QUEUE_REDIS_KEY: null
+PROMETHEUS_ALERT_OCIM_QUEUE_SIZE_THRESHOLD: null
+
```
lgp171188, line 40: Can these be defined elsewhere? I'm asking because the rest of this file only has generic stuff for applications which are not specific to OpenCraft like Ocim is.

playbooks/group_vars/prometheus/public.yml
```diff
@@ -161,7 +164,17 @@ prometheus_rules:
...
+    rules:
+    - alert: "ocim_queue_too_big"
+      expr: 'redis_key_size{key="{{ OCIM_QUEUE_REDIS_KEY }}"} > {{ PROMETHEUS_ALERT_OCIM_QUEUE_SIZE_THRESHOLD }}'
+      for: "0s"
+      labels:
+        severity: "critical"
+        environment: "[[ $labels.environment ]]"
+      annotations:
+        summary: "Ocim queue overloaded."
+        description: "Ocim queue overloaded."
```
lgp171188, line 177: Similar to https://github.com/open-craft/ansible-playbooks/pull/177/files#r487199299

playbooks/roles/redis-exporter/README.md
```diff
@@ -0,0 +1,14 @@
+redis-exporter
+==============
+
+Extends prometheus_redis_exporter_role to be used in Ocim.
```
lgp171188, line 4: @shimulch, @0x29a, imho, this role should be a generic redis-exporter role not specific to Ocim and it should be possible to configure and use it for Ocim by overriding the variables that this role provides. CC @giovannicimolin

playbooks/roles/redis-exporter/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Role Variables
+--------------
+
+A description of the settable variables for this role should go here, including any variables that are in defaults/main.yml, vars/main.yml, and any variables that can/should be set via parameters to the role. Any variables that are read from other roles and/or the global scope (ie. hostvars, group vars, etc.) should be mentioned here as well.
```
lgp171188, line 9: Remove boilerplate text and add actual details about the role variables or add a note to check `defaults/main.yml`.

playbooks/roles/redis-exporter/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Dependencies
+------------
+
+prometheus_redis_exporter_role
```
lgp171188, line 14: @shimulch, @demid, can you add a link to the role repository?

lgp171188: Also depends on the `nginx-proxy` role.

playbooks/group_vars/prometheus/public.yml
0x29a: @lgp171188 while I got you point with moving this elsewhere, I can't find a good way to move this rule.

Would something like assigning anchors to rules, and then overriding `prometheus_rules` with adding of a new rule in `host_vars` acceptable?

giovannicimolin: @0x29a @lgp171188 I don't think we need to remove the rule from here, this is the main alert configuration place. 
We eventually want to add other monitors here (`ocim-exporter`, `edxapp-exporter`), so it doesn't make much sense to keep this too generic, specially since it's a monitoring system specific to our infrastructure.

giovannicimolin: The alternative option would be to move all the rules to `ansible-secrets` but they are not really secret right?

lgp171188: @giovannicimolin, @0x29a, I am okay with leaving this as-is without trying to move it somewhere else to keep this generic because it is not straightforward and doesn't give a lot of benefit. 👍🏼

lgp171188 (approved):
:+1: 
- [ ] I will test and confirm that this works before merging this PR.
- [x] I read through the code

playbooks/ocim.yml
```diff
@@ -9,3 +9,9 @@
 
     - role: ocim
       tags: 'ocim'
+
+    - role: redis-exporter
+      tags: 'redis-exporter'
+      REDIS_EXPORTER_CHECK_KEYS: "{{ OCIM_QUEUE_REDIS_KEY }}"
+      REDIS_EXPORTER_REDIS_ADDR: "{{ OCIM_REDIS_ADDRESS }}"
```
lgp171188, line 16: @0x29a, `OCIM_REDIS_ADDRESS` is missing in the role default variables file.

lgp171188 (approved):
:+1: 
- [x] I tested this along with https://github.com/open-craft/ansible-secrets/pull/200 and verified that the changes work okay
- [x] I read through the code
- [x] Includes documentation

playbooks/ocim.yml
0x29a: @lgp171188, do you mean `playbooks/roles/redis-exporter/defaults/main.yml`? If so, it shouldn't be there since we're trying to make this role generic.

PR #178, opened 2020-08-18
SE-3055: Make redirects temporary

Kelketek:
Change maintenance redirects to use 307 (temporarily moved) rather than 301 (permanently moved) or 302 ('Found').

### DIFF
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -101,16 +101,16 @@ backend error
 # or unscheduled maintenance. The instructions for how to activate them can be
 # found here: https://gitlab.com/opencraft/documentation/private/blob/master/ops/incidents.md
 backend scheduled-maintenance
-    http-response redirect location / unless { capture.req.uri / }
+    http-response redirect location / code 307 unless { capture.req.uri / }
     server scheduled-maintenance-page 127.0.0.1:{{! maintenance_scheduled_server_port !}}
 
 backend unscheduled-maintenance
-    http-response redirect location / unless { capture.req.uri / }
+    http-response redirect location / code 307 unless { capture.req.uri / }
     server unscheduled-maintenance-page 127.0.0.1:{{! maintenance_unscheduled_server_port !}}
 
 # This backend can be used when an instance is being provisioned
 backend provisioning
-    http-response redirect location / unless { capture.req.uri / }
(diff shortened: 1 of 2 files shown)


playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
```diff
@@ -101,16 +101,16 @@ backend error
 # or unscheduled maintenance. The instructions for how to activate them can be
 # found here: https://gitlab.com/opencraft/documentation/private/blob/master/ops/incidents.md
 backend scheduled-maintenance
-    http-response redirect location / unless { capture.req.uri / }
+    http-response redirect location / code 307 unless { capture.req.uri / }
```
lgp171188, line 104: @fox, I don't think these are used at all. And this is not a redirect to an external URL like it happened with `default.opencraft.com` issue. So afaik, there is nothing to fix unless someone can point to something that has to be fixed.

Kelketek: @lgp171188 I don't have access to the repos here-- are you suggesting that maybe nothing needs to be done for this task at all and it should just be closed?

Before we do that, would you be able to run a maintenance page deployment on a test instance on stage to verify we don't see a permanent redirect? I think that would allow us to verify it's unneeded, yes?

lgp171188: @Kelketek, both sites mentioned in the [SE-3055](https://tasks.opencraft.com/browse/SE-3055) task are hosted on the same server. In a recent change, a 301 redirect from HTTP to HTTPS was added to these sites [here](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/wordpress/templates/maintenance-nginx.j2#L5) and [here](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/wordpress/templates/default-nginx.j2). However, these sites are not meant to be accessed directly at all and are only proxied to by our load balancers.

So to avoid the HTTP to HTTPS redirect set up on these virtual hosts, the load balancers will have to directly proxy to the HTTPS URL and this is already done for the default.opencraft.com backend server.

As per the documentation in https://gitlab.com/opencraft/documentation/private/blob/master/ops/incidents.md, the `maintenance.opencraft.com` site is not used for maintenances and instead a local nginx server is used. So the below change to the `maintenace-nginx.j2` virtual host configuration is not needed.

I will have to spend some time to try setting up a maintenance page and then examine what happens.

PR #179, opened 2020-08-28
Fix issues with the load-balancer-v2 role

mtyaka:
Some tasks had to be moved around:

1. Directories under `/etc` need to be created early, before the first task that depends on them
2. The `maintenance.yml` tasks need to run before the tasks from the certbot role. The cerbot role needs a running haproxy (to proxy requests to local nginx), but haproxy won't start if the `maint.map` file created by `maintenance.yml` doesn't exist.

**Testing**:

I don't see any way to properly test this without running it on a fresh server, which is a lot of trouble :/ Since the changes are minor (just changing the order of the tasks), I suppose we could skip a full manual test.

**Reviewers**:

- [x] @swalladge

### DIFF
--- a/playbooks/roles/load-balancer-v2/tasks/main.yml
+++ b/playbooks/roles/load-balancer-v2/tasks/main.yml
@@ -25,6 +25,18 @@
     - ssl-cert
     - jq
 
+- name: Create directories
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: "0700"
+  with_items:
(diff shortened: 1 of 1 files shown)


swalladge:
@mtyaka :+1:

- [x] I tested this: verified that the only changes were moving tasks earlier in the file
- [x] I read through the code
- [ ] NA I checked for accessibility issues
- [x] Includes documentation

swalladge (approved):

PR #180, opened 2020-09-03
[BB-1716] Enable node-exporter explicitly

aayushagra:
This PR explicitly enables node-exporter during provisioning. It fixes an issue where sometimes node-exporter is active but not enabled and hence doesn't restart automatically on reboot.

**JIRA tickets**: 

**Discussions**: 

**Dependencies**: None

**Screenshots**: 

**Sandbox URL**: 

**Merge deadline**: None

**Testing instructions**:

1. Setup a virtualenv and install all requirements
2. Create a fresh ubuntu 16.04 vagrant box
3. Run any provisioning script (such as sprints.yml) through ansible
4. Use 'vagrant ssh' and 'systemctl status nodeexporter' to confirm node-exporter is working
5. Use 'vagrant reload'
6. Use 'vagrant ssh' and 'systemctl status nodeexporter' to confirm node-exporter is working

1. I believe this should fix the problem, and i tried multiple failure situations where without this code node-exporter is not running but with this code node-exporter is running. But node-exporter seems to be auto-disabled if it's not able to contact consul in which case this would not fix the problem (But since a similar approach was tried by @Agrendalath and it successfully fixed the issue with the sprints app, that's unlikely to be the case)

**Reviewers**
- [x] @giovannicimolin 

```

### DIFF
--- a/playbooks/roles/node-exporter/tasks/main.yml
+++ b/playbooks/roles/node-exporter/tasks/main.yml
@@ -11,9 +11,6 @@
   notify:
     - reload consul
 
-- name: flush handlers
-  meta: flush_handlers
-
 - name: enable nodeexporter
   systemd:
     name: nodeexporter.service
--- a/playbooks/roles/node-exporter/tasks/main.yml
+++ b/playbooks/roles/node-exporter/tasks/main.yml
@@ -10,3 +10,13 @@
     dest: /etc/consul/node-exporter.json
   notify:
     - reload consul
+
+- name: flush handlers
+  meta: flush_handlers
+
+- name: enable nodeexporter
+  systemd:
+    name: nodeexporter.service
+    enabled: yes
+    state: started
+    daemon_reload: yes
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: playbooks/roles/node-exporter/tasks/main.yml [hunk 1]
```diff
@@ -10,3 +10,13 @@
     dest: /etc/consul/node-exporter.json
   notify:
     - reload consul
+
+- name: flush handlers
```
giovannicimolin, line 14: @aayushagra Flushing handlers here will force all handlers to be run, instead of at the end of playbooks, which might, in certain cases, break deployments. 

Is there any reason why this is here?
If so, can you run specific actions on which handler you need run instead of flushing everything before the end of the playbooks?

aayushagra: @giovannicimolin Actually i just did that because sometimes node-exporter gets disabled right after consul is reloaded. I guess it's an extra bug that i found and fixed but removing it should still allow the original code to function.

If flush handlers can cause problems it's best to remove it entirely so I'll do that

giovannicimolin (approved):
:+1:

- [ ] I tested this: Did not test the changes locally due to setup issues. Approving since changes are simple and were already tested in the `sprints` app. 
- [x] I read through the code

PR #181, opened 2020-09-08
[SE-3182] Install certbot from snap.

mtyaka:
This updates the certbot role to install certbot from snap instead of the now no longer supported PPA method.
When installed form snap, certbot is automatically kept updated to the latest stable version (currently 1.8.0).

Certbot from snap comes with a systemd timer which periodically runs the `certbot renew` command to renew certificates. However we want to get a notification email when renewing any of the certificates fails, which the systemd timer does not support. We disable the timer and instead add a custom cronfile that runs `certbot renew` twice per day and sends any error output to ops email.

**Test instructions**

This change was deployed to `cert-manager-stage.net.opencraft.hosting` together with https://github.com/open-craft/cert-manager/pull/16. Follow the test instructions on https://github.com/open-craft/cert-manager/pull/16 to verify that requesting a new certificate for an Ocim instance works correctly.

Testing certificate renewal is trickier since certificates are only automatically renewed when they near their expiration date.
You can run `/snap/bin/certbot renew --dry-run` to see that the command works, but you will see some legitimate errors due to DNS not being set up for some of the instances on Ocim stage.

You can inspect `/var/log/syslog` to verify that the certbot cronjob runs twice per day and you can examine the files under `/var/log/letsencrypt/` to check the output. 

**Reviewers**:

- [ ] @Kelketek
- [ ] @lgp171188


lgp171188:
@mtyaka, how do we roll out this change to the non-cert-manager servers? We may need ad-hoc tasks to remove the `certbot` PPA package before installing the snap package.

CC @Kelketek

mtyaka:
> @mtyaka, how do we roll out this change to the non-cert-manager servers? We may need ad-hoc tasks to remove the `certbot` PPA package before installing the snap package.

@lgp171188 The changes in this PR include [a task that removes the PPA package if it exist|https://github.com/open-craft/ansible-playbooks/blob/66b56eeada829e66a9122ef2ba2f876cbe015f62/playbooks/roles/certbot/tasks/main.yml#L32-L38].

playbooks/roles/certbot/tasks/main.yml
```diff
@@ -54,6 +61,33 @@
...
+- name: Create cron job to periodically renew certificates
+  cron:
+    name: certbot renew
+    user: root
+    job: "perl -e 'sleep int(rand(43200))' && /snap/bin/certbot -q renew"
+    minute: '0'
+    hour: '*/12'
+    cron_file: certbot-renew
+
+- name: Set MAILTO variable for certbot renew cron job
```
lgp171188, line 83: @mtyaka, since this will be useful only in production and sending these emails from other environments to the Ops list adds to the email clutter, can we make this conditional?

playbooks/roles/certbot/tasks/main.yml
```diff
@@ -29,14 +29,21 @@
...
 
-- name: add certbot PPA
-  apt_repository:
-    repo: ppa:certbot/certbot
+- name: Remove certbot apt package if installed
+  apt:
+    name: "{{ item }}"
+    state: absent
+  with_items:
+    - python-certbot-nginx
```
lgp171188, line 37: @mtyaka, do we know for sure that the functionality provided by this package is either included in the `certbot` snap package or that we don't need/use it?

playbooks/roles/certbot/tasks/main.yml
Kelketek, line 83: Have we tested to make sure this works? I'm not able to see the ops@ email. Can we force a failure?

lgp171188 (approved):
@mtyaka, :+1: unless @Kelketek has any additional review comments, you can merge this PR after answering/addressing the 2 review comments.

- [x] I tested this on the stage `cert-manager` server where this has already been deployed and verified that `certbot` apt package is removed and the snap package is installed, the certificate provisioning, renewal and deployment to consul via the post renew hook all work okay.
- [x] I read through the code

Kelketek (commented):
As long as we're sure the email works, and you've addressed @lgp171188 's concerns, it's 👍 from me. 

- [X] I tested that provisioning a new instance with this configuration issued and installed a valid (if not verified, since it's a test environment) SSL certificate.
- [X] I read through the code.

playbooks/roles/certbot/tasks/main.yml
lgp171188: @Kelketek, we can force a failure by removing the DNS records for an instance whose certificates we want to try renewing.

lgp171188: @Kelketek, @mtyaka, I tried to force a renewal failure by removing the DNS records for the domains in a certificate and attempted to force renew it. But that didn't work because certbot is caching the domain validations from a server for a certificate for 60 days. So I could only check and verify that emails are sent to the `MAILTO` address for any stdout/stderr messages in command output.

playbooks/roles/certbot/tasks/main.yml
mtyaka: That package contains scripts that will automatically update nginx configuration files, which we don't use.

playbooks/roles/certbot/tasks/main.yml
mtyaka: Yup, good idea. I'll do that.

mtyaka: I did get an email to `MAILTO` when I was still testing with my own email (before I switched to ops@). It was not exactly because certificate renewal failed, but it was because the cryptography library printed some python 3.5 incompatibility warnings.

mtyaka: @Kelketek @lgp171188 If you're subscribed to the ops list, you should have received an email alert with the subject "[ops] Cron <root@cert-manager-stage> perl -e 'sleep int(rand(43200))' && /snap/bin/certbot -q renew" a few minutes ago.

lgp171188: @mtyaka, I am not subscribed to that list. But if you are and have received the email, that is good enough for me 👍🏼

lgp171188: @mtyaka, did you get a chance to address this?

playbooks/roles/certbot/tasks/main.yml
```diff
@@ -29,14 +29,21 @@
...
-- name: install certbot apt package
+- name: Install snapd
   apt:
-    update_cache: yes
-    name: python-certbot-nginx
+    name: snapd
+
+# TODO: We should switch to the snap ansible module after we upgrade to Ansible 2.8 or newer.
+- name: Install certbot snap
+  command: snap install certbot --classic
```
lgp171188, line 46: @mtyaka, can we add a comment to explain why we are removing the apt package and installing the snap package?

lgp171188 (approved):
@mtyaka, :+1: if @Kelketek has no other concerns, you can merge the PR after addressing the 2 open comments of mine.

- [x] I tested this on cert-manager-stage and verified that it works okay
- [x] I read through the code
- [ ] Includes documentation

playbooks/roles/certbot/tasks/main.yml
mtyaka: Not yet @lgp171188, I'll take care of it later this week.

Kelketek:
Just wanted to prod this PR and confirm that it's good to go from my end, in case that wasn't clear. 👍

Kelketek:
Confirming that the new changes look good to me, too! 👍

playbooks/roles/load-balancer-v2/tasks/main.yml
```diff
@@ -154,27 +154,12 @@
...
-    # We shouldn't need to pass any parameters to "certbot renew", since it
-    # reuses the parameters used when a certificate was issued for the last time.
-    # However, not passing in the port results in letsencrypt complaining about
-    # port 443 being in use.  (Seems to be a bug in certbot.)
-    job: certbot renew --preferred-challenges http-01 --http-01-port 8080 -q
-    hour: "*/12"
-    minute: 42
-    cron_file: letsencrypt-renew
-    user: root
-
```
lgp171188, line 177: @mtyaka, is it safe to remove this? Iirc, the ACME verification requests on the default HTTP port are all proxied to the {{cert-manager}} server and this custom port `8080` is used for validating the SSL certificate for the `<haproxy hostname>` certificate. Removing this will prevent those certificates from being renewed.

Does the default cron job automatically renew this certificate using a custom port as well? While running `/snap/bin/certbot certificates` lists the certificate for the haproxy host name, I don't see anything in the renewal configuration to indicate that a separate port has to be used for the validation.

lgp171188 (changes requested):
@mtyaka, we have to set the `certbot` path in the `cert-manager.sh` script so that the code [here](https://github.com/open-craft/cert-manager/pull/16/files#diff-4bd6ead4fa425b7b5cdeaa34e60cf3f8R57) can use it.

playbooks/roles/certbot/tasks/main.yml
lgp171188: Resolved [here])https://github.com/open-craft/ansible-playbooks/pull/181/files#diff-a664225e5a176729b75a4704ef78d6a6R32)

playbooks/roles/certbot/tasks/main.yml
lgp171188: Resolved [here](https://github.com/open-craft/ansible-playbooks/pull/181/files#diff-a664225e5a176729b75a4704ef78d6a6R89)

playbooks/roles/sanity-checker/files/sanity_check.py
```diff
@@ -99,7 +99,10 @@ def get_dev_mounts():
         parts = line.split()
         mount = Mount(parts[1], parts[0])
         if mount.device.startswith("/dev"):
-            results.append(mount)
+            # /dev/loopX devices are mounted images that are expected to be 100% full,
+            # so we don't want to monitor those. Snopd uses these extensively.
```
lgp171188, line 103:
```suggestion
            # so we don't want to monitor those. Snapd uses these extensively.
```

lgp171188 (changes requested):
@mtyaka, I have left a couple of comments.

playbooks/roles/certbot/tasks/main.yml
```diff
@@ -29,14 +29,23 @@
     port: 80
     proto: tcp
 
-- name: add certbot PPA
-  apt_repository:
-    repo: ppa:certbot/certbot
+# Certbot apt packages are outdated and no longer officially supported by certbot.
+# The oficially supported installation method is via snap.
+- name: Remove certbot apt package if installed
```
lgp171188, line 34: @mtyaka, it looks like this doesn't remove the dummy cronjob that the `certbot` package installs. While the cronjob itself will do nothing, it will be a good idea to remove it to avoid any confusion.

playbooks/roles/load-balancer-v2/tasks/main.yml
mtyaka: @lgp171188 

> Iirc, the ACME verification requests on the default HTTP port are all proxied to the {{cert-manager}} server and this custom port 8080 is used for validating the SSL certificate for the <haproxy hostname> certificate.

It used to be like that in the past, but now requests for the haproxy certificates itself are [also proxied via haproxy|https://github.com/open-craft/ansible-playbooks/blob/8b0ba942993a70d255054821ea643aac39199195/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl#L51] so that is no longer needed. This cronjob was not working anymore, it was failing with "Problem binding to port 8080: Could not bind to IPv4 or IPv6.. Skipping.".

> While running /snap/bin/certbot certificates lists the certificate for the haproxy host name, I don't see anything in the renewal configuration to indicate that a separate port has to be used for the validation.

I manually removed the port parameters from certificate renewal configurations where they were still present since they're no longer needed as mentioned above. I verified renewals with updated configurations were working with `certbot renew --dry-run. I can also see that the certificates on haproxy-stage-* servers got successfully renewed with the new cronjob yesterday and settings during the night.

mtyaka:
> @mtyaka, we have to set the `certbot` path in the `cert-manager.sh` script so that the code [here](https://github.com/open-craft/cert-manager/pull/16/files#diff-4bd6ead4fa425b7b5cdeaa34e60cf3f8R57) can use it.

I made the setting default to `/snap/bin/certbot` in both [manage_certs.py](https://github.com/open-craft/cert-manager/blob/ed84f3cdc5f8650c80fdef0b2be6c555c654672d/manage_certs.py#L48) and [deploy_cert.py](https://github.com/open-craft/cert-manager/blob/ed84f3cdc5f8650c80fdef0b2be6c555c654672d/deploy_cert.py#L26) so that we don't have to pass the path in every place we invoke these scripts.

playbooks/roles/sanity-checker/files/sanity_check.py
mtyaka: Thanks for catching this; I fixed it.

playbooks/roles/certbot/tasks/main.yml
mtyaka: Good idea 👍 I'll do it manually.

mtyaka: Manually removed the old cronjob from all servers to which we already installed the new version.

playbooks/roles/load-balancer-v2/tasks/main.yml
lgp171188: Thank you for the details @mtyaka 👍🏼

playbooks/roles/certbot/tasks/main.yml
lgp171188: @mtyaka, the `certbot` package hasn't been removed properly on one of our servers and I have mentioned the details in my reply on the task. Can you check that?

lgp171188 (approved):
:+1: 
- [x] I tested the changes deployed using this PR and verified that everything works okay
- [x] I read through the code
- [x] Includes documentation in the form of inline comments

playbooks/roles/certbot/tasks/main.yml
mtyaka: Thanks for catching that... I'm not sure what happened, but I do remember that some ansible tasks were taking unusually long to complete on that server when I ran the playbooks yesterday. Maybe 

I re-ran the playbook today and it was stuck on the "Remove certbot apt package if installed" for almost 20 minutes. However after 20 minutes it successfully completed and I verified that the certbot apt package is no longer installed on the server.

PR #184, opened 2020-09-28
[BB-3018] Load balancers should serve the maintenance pages from the local nginx vhosts and not redirect to external domains

shimulch:
This PR fixes the wrong maintenance page issue during scheduled or unscheduled maintenance and during provisioning.

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-3018

~~**Discussions**:~~
 
**Dependencies**: None

~~**Screenshots**: ~~
~~**Sandbox URL**:~~

**Merge deadline**: "None"

**Testing instructions**:
1. Move an instance to maintenance mode, proper maintenance page should show
2. Provisioning page should show from local Nginx instead of using external server IP

~~**Author notes and concerns**:~~

**Reviewers**
- [ ] @lgp171188 

~~**Settings**~~

### DIFF
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -119,6 +119,7 @@ backend provisioning
   {{- with $config := .Value | parseJSON }}
     {{- if $config }}
     {{ $health_check := $config.health_checks_enabled }}
+    {{- if $config.active_app_servers }}
 # Backend configuration for {{ $config.domain }}
 backend be-{{$config.domain_slug}}
     cookie openedx-backend insert postonly indirect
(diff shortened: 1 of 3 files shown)


lgp171188 (commented):
@shimulch, these changes look okay to me but will have to be tested before approval.

playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
```diff
@@ -119,6 +119,7 @@ backend provisioning
   {{- with $config := .Value | parseJSON }}
     {{- if $config }}
     {{ $health_check := $config.health_checks_enabled }}
+    {{- if $config.active_app_servers }}
```
lgp171188, line 122: @shimulch, this condition can be combined into the existing if condition with `and` ([docs](https://golang.org/pkg/text/template/#hdr-Functions))

shimulch: @lgp171188 Thanks. Updated.

lgp171188 (approved):
:+1: 
- [x] I tested these changes in the stage environment and verified that:
  * When an instance doesn't have active appservers, the provisioning page from the local nginx server is served.
  * When an instance has active appservers, a proper backend configuration is generated and the instance works okay.
  * Both the maintenance backends now show the correct page instead of the error page.
- [ ] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation

PR #186, opened 2020-10-06
[BB-3036] Added new PPA for MySQL 5.6 support

farhaanbukhsh:
The old PPA we have added to our xenial instance was gone,
this PPA is coming from ubuntu's earlier releases where MySQL 5.6
was supported.

**JIRA tickets**: [BB-3036](https://tasks.opencraft.com/browse/BB-3036)

~~**Discussions**: Link to any public dicussions about this PR or the design/architecture. Otherwise omit this.~~

~~**Dependencies**: None~~

~~**Screenshots**: Always include screenshots if there is any change to the UI.~~

~~**Sandbox URL**: TBD - sandbox is being provisioned.~~

~~**Merge deadline**: "None" if there's no rush, "ASAP" if it's critical, or provide a specific date if there is one.~~

**Testing instructions**:

1. Make a directory for this vargrant box, "mkdir ansible_exp" and "cd ansible_exp"
2.  Do `vagrant init ubuntu/xenial64`
~3. `vagrant init`~
4.  Clone this repot and switch branch to `farhaan/bb-3036-fix-mysql-deployment-playbook` inside this make sure to run ``git submodule update --init --recursive`` to get the modules.
5.  We need to add sync_folder in vagrant file, do vim Vagrantfile and add the following line 
``
config.vm.synced_folder "/path-to/your/ansible-playbooks", "/ansible-playbooks"
`` 
Here add replace the first parameter with your cloned repo path
and edit the files as mentioned in the diff below

[Ansible Diff](https://gist.github.com/farhaanbukhsh/890a1cb8d74a3ac790efc8319d49f80a)

and create an inventory file you can do `vim inventory` and put 

```
[mysql]
127.0.0.1 ansible_connection=local

```
in the file

6. ``vagrant up``
7. ``vagrant ssh``
8. ``cd  /ansible-playbook``
9. If you are using a machine with low spec you can uncomment the ``mysql_innodb_buffer_pool_size`` variable else it should be fine
10. sudo apt update  &&  sudo apt upgrade
11. sudo apt install python3 python3-pip build-essential libssl-dev libffi-dev python-dev -y
12. export LC_ALL=C
13.  pip3 install -r requirements.txt
14. I am going to run two roles separately for easy debugging you can run in one shot too
15.`` cd  playbooks``
16. ``ANSIBLE_RETRY_FILES_ENABLED=0 ansible-playbook -vvv mysql.yml --tags 'mysql' -i ../inventory``
17. ``ANSIBLE_RETRY_FILES_ENABLED=0 ansible-playbook -vvv mysql.yml --tags 'geerlingguy.mysql' -i ../inventory``
18. Above both commands should go clean and if you do `mysql --version` command you should see `5.6.16`
19. Running ``systemctl status mysql.service`` should also show you that it's active. 

This enables the old trusty thar repo, I think this is fine since we are anyhow using outdated versions,
I feel this method should work on 20.04 as well if I am not wrong I haven't tested that yet though.

**Reviewers**
- [ ] @giovannicimolin


giovannicimolin (changes requested):
This doesn't work with Ubuntu 18.04 and 20.04, as a lot of dependencies are not installable anymore.

You need to download the package and install it manually.
I've found this: https://askubuntu.com/questions/1029634/install-mysql-5-6-on-ubuntu-18-04-lts

You can also look at Ansible Galaxy and check if someone already did this.

playbooks/roles/mysql/tasks/main.yml
```diff
@@ -3,7 +3,7 @@
 # edX is currently using MySQL 5.6, which is not available on Xenial by default.
 - name: Add PPA for MySQL 5.6
   apt_repository:
-    repo: "ppa:ondrej/mysql-5.6"
+    repo: "deb http://archive.ubuntu.com/ubuntu trusty universe"
```
giovannicimolin, line 6: @farhaanbukhsh This repo will not work unfortunately.
Our servers use MySQL 5.6.35, and we cannot downgrade minor versions.

playbooks/group_vars/mysql/public.yml
```diff
@@ -9,8 +9,10 @@ mysql_table_open_cache: 200000
 mysql_wait_timeout: 300
 mysql_skip_name_resolve: true
 mysql_packages:
-  - mysql-server-5.5
-  - mysql-client-5.5
+  - mysql-server-5.6
+  - mysql-client-5.6
+# Incase test has to be done on a small machine
+# mysql_innodb_buffer_pool_size: 20M
```
giovannicimolin, line 15: Remove these lines.
```suggestion
```

playbooks/roles/mysql/tasks/main.yml
giovannicimolin: @farhaanbukhsh Can you look at another PPA or Ansible Galaxy playbook to install the right MySQL version? An upgrade is fine, but a downgrade is not.

giovannicimolin (changes requested):

playbooks/roles/mysql/tasks/main.yml
farhaanbukhsh: sure I a having a look, mostly trusty stopped at 5.6.33 and there are not there roles that I could see supporting 5.6, most of the people were using same PPA as us and as it stopped they stopped supporting the version. But I am trying to dig in and see if I can find a PPA or a way to get 5.6.35 or above.

giovannicimolin (changes requested):
@farhaanbukhsh The testing instructions are not working for me. I'll try a bit more, can you check and make sure they work?

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Tested the role on a local vagrant VM.
2. Checked that the correct MySQL version was installed.
- [x] I read through the code

@farhaanbukhsh Managed to get it working from a clean VM. Nice work here! Really clean solution.

playbooks/roles/mysql/defaults/main.yml
```diff
@@ -39,3 +39,19 @@ MYSQL_EXPORTER_CONSUL_SERVICE:
       - monitor-https
     port: 19104
     enable_tag_override: true
+
+# MySQL Binaries for pinning to version 5.6.35 #########################################
+
+MYSQL_DEBIAN_SNAPSHOT_URL: http://snapshot.debian.org/archive/debian/20170120T212942Z/pool/main/m/mysql-5.6
```
lgp171188, line 45: @farhaanbukhsh, @giovannicimolin, is this a supported and maintained debian package? If not, we should just use the binaries from MySQL directly.

farhaanbukhsh: These are signed and supported binaries from Debian. @lgp171188

lgp171188: @farhaanbukhsh, signed by Debian doesn't matter. "Supported" means being a supported package in a supported release, which is "Debian 10" and "Debian 9" LTS (but may not have a supported MySQL package).

lgp171188: @giovannicimolin ^

farhaanbukhsh: Okay that does make sense, should we schedule a task to do that?

farhaanbukhsh: @lgp171188 I could see only https://downloads.mysql.com/archives/community/?version=5.6.44 5.6.44 which has Ubuntu supported binaries. Is there any other place where we can get deb packages for 5.6.50?

lgp171188: @farhaanbukhsh, the task description says

> Update https://github.com/open-craft/ansible-playbooks/tree/master/playbooks/roles/mysql so that MySQL is installed from the binaries.

It doesn't say "deb" packages, though they would be very convenient and easy to use. That is also why I was referring to installing from the binary tarball release from the MySQL site.

I'll let @giovannicimolin comment on this and decide on the next steps, if any.

farhaanbukhsh: @lgp171188 ,

There is more clarification in the comment https://tasks.opencraft.com/browse/BB-3036?focusedCommentId=169392&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-169392 

Also I think it was a call because of MySQL 5.6 reaching EOL and a lot of services migrating and using 5.7 even for Edx. 😄 
 So we decided to have a workable solution till then.

giovannicimolin: @lgp171188 I think I've decided for the approach because of two reasons:

1. MySQL 5.6 EOL: which would force us to upgrade in the near future.
2. Koa and upcoming releases with update MySQL version compatibility.

So it didn't make sense to spend a lot of hours given we'll have to rework our playbooks to deploy the newest version soon. I just wanted the playbooks to work in case we need to do a emergency redeployment until we upgrade the service.

PR #187, opened 2020-10-12
SE-3328 Add newrelic infra and rabbit integration

swalladge:
This adds support for newrelic infrastructure plugin with the rabbitmq integration running on rabbitmq instances.  It should allow us to collect metrics on rabbitmq, and use those metrics for alerting (for example, if queue lengths go through the roof).

It also fixes some issues with running the playbook.

**Test instructions**:

- deploy rabbitmq and verify that it completes successfully
- check newrelic and verify that it is receiving metrics from the rabbitmq instance

### DIFF
--- a/playbooks/roles/rabbitmq/meta/main.yml
+++ b/playbooks/roles/rabbitmq/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+  - role: newrelic_infra
--- a/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/playbooks/roles/rabbitmq/tasks/main.yml
@@ -163,23 +163,10 @@
     user: guest
     state: absent
 
-- name: add rabbitmq vhost
-  rabbitmq_vhost:
-    name: "{{ RABBITMQ_VHOST }}"
-    node: "rabbit@{{ RABBITMQ_HOSTNAME }}"
-    state: present
-
-- name: create rabbitmq user
-  rabbitmq_user:
-    user: "{{ RABBITMQ_USER }}"
-    password: "{{ RABBITMQ_PASS }}"
-    vhost: "{{ RABBITMQ_VHOST }}"
-    state: present
-
+# TODO: what is this used for?
 - name: ensure vhost /test is present
   rabbitmq_vhost:
     name: /test
-    node: "rabbit@{{ RABBITMQ_HOSTNAME }}"
     state: present
 
 - name: ensure newrelic rabbitmq monitoring is present
--- a/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/playbooks/roles/rabbitmq/tasks/main.yml
@@ -181,3 +181,27 @@
     name: /test
     node: "rabbit@{{ RABBITMQ_HOSTNAME }}"
     state: present
+
+- name: ensure newrelic rabbitmq monitoring is present
+  apt:
+    pkg: "nri-rabbitmq"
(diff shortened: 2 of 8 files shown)


playbooks/roles/rabbitmq/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - role: newrelic_infra
```
pomegranited, line 3: Should also update these either in [rabbitmq/defaults/main.yml](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/roles/rabbitmq/defaults/main.yml#L24-L25) or in the private vars, since that's the version that's now running on stage:

```
ERLANG_VERSION: "1:23.1"   # not sure about this format?
RABBITMQ_VERSION: "3.8.9"
```

playbooks/roles/rabbitmq/templates/rabbitmq-config-newrelic-infra.yml.j2
```diff
@@ -0,0 +1,21 @@
...
+# This file is managed via Ansible.  Manual changes to this file will be overwritten.
+################
+integration_name: com.newrelic.rabbitmq
+
+instances:
+  - name: "{{ RABBITMQ_HOSTNAME }}"
+    command: all
+    arguments:
+      hostname: "{{ RABBITMQ_HOSTNAME }}"
+      port: 15671
```
pomegranited, line 11: Please use the [`RABBITMQ_HTTPS_PORT`](https://github.com/open-craft/ansible-playbooks/blob/e3af5face67ea9071f1de5888c0aa2f7a10e7310/playbooks/roles/rabbitmq/defaults/main.yml#L4-L5) variable instead.

pomegranited (changes requested):
I'm testing the updated Stage rabbitmq with a new Ocim instance, so will give a formal +1 once that's running ok.

playbooks/roles/rabbitmq/meta/main.yml
swalladge: That's part of the reason we're getting dependency issues on stage; I think we'll need to re-pin back... Will have a play and see how it goes.

swalladge: As discussed, we don't want to deal with upgrading rabbitmq, so the existing pinned versions can stay. :)

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -163,21 +163,32 @@
     user: guest
     state: absent
 
-- name: add rabbitmq vhost
+# TODO: what is this used for?
+- name: ensure vhost /test is present
   rabbitmq_vhost:
-    name: "{{ RABBITMQ_VHOST }}"
-    node: "rabbit@{{ RABBITMQ_HOSTNAME }}"
+    name: /test
```
pomegranited, line 169: Could just be a basic test of rabbitmq's functionality.. I think it's ok to leave in.

pomegranited (approved):
:+1:

- [x] I tested this by verifying that newrelic agent is running on rabbitmq-stage, and stage is still running the same version of rabbitmq as prod.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> N/A
- [x] I made sure any change in configuration variables is reflected in the ansible-secrets repo.

PR #190, opened 2020-11-09
[BB-3156] Update Sprints README

Agrendalath:
This:
1. Adds a short guide to deploying Sprints on smaller instances. This was not added to playbooks, because the approach might differ depending on the selected hosting provider (some providers offer 20GB as the lowest possible storage size, so it would not be a problem there) and a chosen instance flavor (which can be changed in case of a required redeployment).
2. Removes TODO items that have been fixed in #133.


swalladge (approved):
@Agrendalath :+1:

- [ ] I tested this: checked that the instructions matched reality
- [x] I read through the code
- [x] Includes documentation

PR #191, opened 2020-11-12
BB-3167 Remove mmapv1 engine configuration

viadanna:
Description
=========

This PR removes a dangling configuration that sets the MongoDB engine to mmapv1, which is not what we want and shouldn't be set here.

Testing
======

Check related https://github.com/open-craft/ansible-secrets/pull/219


0x29a (approved):
:+1:

- [x] I read through the code

PR #192, opened 2020-11-18
[SE-3417] Remove unused MYSQL_OPENSTACK_DB_DEVICE variable.

mtyaka:
It's not used anywhere in the playbook.

**Reviewers**:

- [ ] @mavidser


mavidser (approved):
:+1: Thanks for the cleanup @mtyaka - verified that the variable isn't used anywhere.

- [x] I tested this: Relied on @mtyaka's tests, and checked git history for when the usage of the variable was removed
- [x] I read through the code
- N/A I checked for accessibility issues
- [x] Includes documentation

PR #193, opened 2020-11-24
[SE-3417] Use certbot from snap in common-apache2 role.

mtyaka:
This port the relevant parts from the [certbot role](https://github.com/open-craft/ansible-playbooks/tree/master/playbooks/roles/certbot) to `common-apache2`.

The most important part is upgrading certbot to the version provided as a snap package, which we also use on the rest of our servers, since the PPA installation method is no longer supported upstream and the version in the PPA is outdated.

**Test instructions**

This was deployed to the three servers that use this role:
- mit-catchpy.opencraft.hosting
- mit-hxat.opencraft.hosting
- mit-loris.opencraft.hosting

The first of the certificates will be up for renewal in about two weeks, so to verify that the renewal process is actually working we'll have to wait.
In order to review this PR, I think it's enough to read through the code make sure it makes sense.

**Reviewers**:
- [ ] @mavidser

### DIFF
--- a/playbooks/roles/common-apache2/tasks/main.yml
+++ b/playbooks/roles/common-apache2/tasks/main.yml
@@ -1,13 +1,7 @@
 ---
 # This role contains common tasks that will run on all nodes.
 
-- name: add `ppa:certbot/certbot` repo
-  become: yes
-  apt_repository:
-    repo: ppa:certbot/certbot
-
 - name: install common packages
-  become: yes
   apt:
     name: "{{ packages }}"
     update_cache: yes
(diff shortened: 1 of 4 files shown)


playbooks/roles/common-apache2/tasks/main.yml
```diff
@@ -1,32 +1,59 @@
...
     - libffi-dev
     - openssl
-    - python-certbot-apache
     - python-dev
     - python-pip
     - python-setuptools
 
+- name: Install snapd
+  apt:
+    name: snapd
```
mavidser, line 21: Can just be added to the list above

mavidser (approved):
LGTM @mtyaka :+1:

- [x] I tested this: Didn't run the playbook, but verified that the servers mentioned have newer certs, which were setup through this role.
- [x] I read through the code

playbooks/roles/common-apache2/tasks/main.yml
mtyaka: I added it to the list initially, but later decided to move it to a distinct task to keep tasks that were "ported" over from the certbot role separated.

mtyaka:
Thanks @mavidser!

PR #194, opened 2020-12-07
backup-pruner: don't modify /var/cache permissions

mtyaka:
The /var/cache folder already exist on Ubuntu (and most other linux distributions), so there's no need to create it in the backup-pruner playbook. In fact forcing `0700` permissions (which is stricter than the default) leads to some periodic system tasks failing.

**Testing**:

This is a trivial change and testing shouldn't be necessary.

### DIFF
--- a/playbooks/roles/backup-pruner/tasks/main.yml
+++ b/playbooks/roles/backup-pruner/tasks/main.yml
@@ -6,12 +6,9 @@
 
 - name: create dirs
   file:
-    name: "{{ item }}"
+    name: /etc/tarsnap
     mode: 0700
     state: directory
-  with_items:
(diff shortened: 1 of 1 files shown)


itsjeyd (approved):
@mtyaka Thanks for these changes, they LGTM :+1:

- [x] I did not test this (trivial change), but confirmed that after [manually fixing permissions](https://tasks.opencraft.com/browse/SE-3417?focusedCommentId=179048&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-179048) of `/var/cache`, notifications about errors we were seeing earlier stopped.
- [x] I read through the code

PR #199, opened 2021-01-26
[SE-3967] fix: invalid HTTP host header issue

gabor-boros:
This PR ensures that client connections with an invalid HTTP host header will be dropped.

**Dependencies**: None

**Screenshots**:

Screenshot

**Sandbox URL**: N/A

**Merge deadline**: "None"

**Testing instructions**:

1. Initiate a connection to https://crafty.opencraft.com
2. Validate that you got a HTTP 404
3. Set the HTTP Host header to something else like `fake.domain` and call the same URL
4. Validate that you got a HTTP 444 response

N/A

**Reviewers**
- [ ] @Agrendalath


Agrendalath:
@gabor-boros, the approach seems to be a little heuristic. As you can see [here](https://sentry.io/organizations/opencraft/issues/2121246063/?project=5572527), it doesn't handle the `http GET 147.135.195.76:5000` case.

I didn't dig into this too much, but it seems that `sentry_sdk` is capturing [django.security.DisallowedHost](https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/config/settings/production.py#L182-186) (it's not a traceback, just an event from the logger). As this event is logged on the `ERROR` level, and sentry has its [event_level](https://docs.sentry.io/platforms/python/guides/logging/#options) set to `logging.ERROR` [here](https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/config/settings/production.py#L197), it might be catching this. I see three potential solutions here:
1. [Ignore this logger](https://docs.sentry.io/platforms/python/guides/logging/#ignoring-a-logger) via the `sentry_sdk`:
   ```python
   from sentry.integrations.logging import ignore_logger
   ignore_logger("django.security.DisallowedHost")
   ```
1. Reduce the [level](https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/config/settings/production.py#L183) of the [django.security.DisallowedHost logger](https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/config/settings/production.py#L182).
1. Set the `event_level` of `sentry_sdk` to `None`. This is what we're doing [in Sprints](https://github.com/open-craft/sprints/blob/master/config/settings/production.py#L208-208), and, if I understand this correctly, the only real implication here is that we cannot log anything to Sentry via `logging.error()` or `logging.exception()`. We're just using [such workaround](https://github.com/open-craft/sprints/blob/master/sprints/dashboard/libs/mattermost.py#L25-L28) for passing messages via exceptions.

Also, we could disable handlers for this logger, as suggested [here](https://docs.djangoproject.com/en/dev/topics/logging/#django-security), to declutter our Docker logs a bit. These errors will appear when you check `docker-compose logs django` and they don't seem to be having any real value.

gabor-boros:
@Agrendalath Although it is a bit strict implementation, allow trying to call the service with different Host headers then it should be called with (even IP addresses) can easily lead to security issues since these requests are usually looking for potential exploits.

In my opinion it would be better to handle the issue (from security point of view), then mitigating it by muting the root cause of the issue.

These logs mostly generated by automated vulnerability checkers.

Agrendalath:
@gabor-boros, yep, I'm not saying that we should remove the host validation.
If we'll manage to set something like a [default_server](https://www.nginx.com/resources/wiki/start/topics/examples/server_blocks/#a-default--22catch-all-22-server-block) in Nginx to block direct IP calls in a reasonable time, without breaking other services, like Consul and node-exporter, then +1 to doing this and leaving the logging as is.

gabor-boros:
@Agrendalath No need for the IP blocking. I did some investigation around the topic.

So, the IP address wasn't served by NGINX at all. It was served by docker-compose. NGINX properly handles the `444` return, though it is not reached at all.

`docker-compose` **changes** the `iptables` when not directly bind to `127.0.0.1`. After `docker-compose.yml` is changed to use `127.0.0.1`, the `iptables` not modified, hence `<IP>:<PORT>` cannot be reached and NGINX serves the requests properly.

So to conclude: There is no need to make restrictions for IP addresses or set `default_server`.

You can test the changes by executing:

```shell
curl \
 --compressed \
 -H "Accept: */*" \
 -H "Accept-Encoding: gzip" \
 -H "Host: 147.135.195.76:5000" \
 -H "User-Agent: Mozilla/5.0 (compatible; CensysInspect/1.1; +https://about.censys.io/)" \
 "http://147.135.195.76:5000/"
```

playbooks/roles/crafty-bot/files/docker-compose.yml
```diff
@@ -32,5 +32,5 @@ services:
     <<: *django
     image: registry.gitlab.com/opencraft/dev/crafty-bot:flower-latest
     ports:
-      - "5555:5555"
+      - "127.0.0.1:5555:5555"
```
Agrendalath, line 35: @gabor-boros, this is going to make Flower inaccessible from outside of the instance, no? We might either open it or add an Nginx config to do a proxy for this.

Agrendalath:
@gabor-boros, should we also add `CELERY_BROKER_URL: redis://redis:6379/0` here?

playbooks/roles/crafty-bot/files/docker-compose.yml
gabor-boros: I'm planning to add an NGINX location for it.

gabor-boros:
> @gabor-boros, should we also add `CELERY_BROKER_URL: redis://redis:6379/0` here?

@Agrendalath Not necessarily. The configuration already defines the `REDIS_URL` at https://github.com/open-craft/ansible-playbooks/blob/b36844bc1ff982db4741b71e7e0d49c9eb03a62d/playbooks/roles/crafty-bot/defaults/main.yml#L57 which is used as the `CELERY_BROKER_URL` in https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/compose/production/django/entrypoint#L10.

In case we want to specify a different broker than Redis or a different DB, we could set it, otherwise, it is not needed.

playbooks/roles/crafty-bot/files/docker-compose.yml
Agrendalath: We should probably add support for [url_prefix|https://flower.readthedocs.io/en/latest/config.html#url-prefix] in [here](https://gitlab.com/opencraft/dev/crafty-bot/-/blob/master/compose/production/django/celery/flower/start#L10) and introduce a new variable to specify it. Currently [https://crafty.opencraft.com/celery-flower/] has broken statics and links.

Agrendalath:
@gabor-boros,
> In case we want to specify a different broker than Redis or a different DB, we could set it, otherwise, it is not needed.

There is one more use-case here - it's in the ticket's description ;)
> we might also consider adding CELERY_BROKER_URL=redis://redis:6379/0 to `.env` file, because without this you cannot get into the shell simply via docker-compose exec django python manage.py shell, but need to do this workaround:
> ```python
> docker-compose exec django sh
> CELERY_BROKER_URL=redis://redis:6379/0 python manage.py shell
> ```

gabor-boros:
@Agrendalath Could you please elaborate on this?
I was able to get the shell, and nothing changed

Screenshot 2021-02-02 at 21 10 17

playbooks/roles/crafty-bot/meta/main.yml
```diff
@@ -6,3 +6,22 @@ dependencies:
     NGINX_PROXY_INTERNAL_PORT: 5000
     NGINX_PROXY_USE_SSL: "{{ USE_SSL }}"
     NGINX_PROXY_SERVER_NAME: "{{ API_HOSTNAME }}"
+    NGINX_PROXY_EXTRA_CONFIG: |
+      # SE-3967 - Prevent Invalid HTTP_HOST headers
+      if ( $host !~* ^{{ all_env_tokens.DJANGO_ALLOWED_HOSTS }}$ ) {
+          return 444;
+      }
+
+      location /flower/ {
```
Agrendalath, line 15: @gabor-boros, I'm not a fan of hardcoding this path, but it can be addressed at a later point, as [this MR](https://gitlab.com/opencraft/dev/crafty-bot/-/merge_requests/4) has already been reviewed by @nizarmah and merged.

Agrendalath (approved):
:+1:

- [x] I tested this: checked that the instance is rejecting requests with a wrong header and Flower works correctly
- [x] I read through the code
- [x] I checked for accessibility issues: n/a
- [x] Includes documentation: n/a

Agrendalath:
@gabor-boros, it seems like there is one more [edge case](https://sentry.io/organizations/opencraft/issues/2121246063/?project=5572527), as we've just bumped into the same issue.

playbooks/roles/crafty-bot/meta/main.yml
nizarmah: Ah sorry about that, I lack the experience with flower to know which practices are preferred and which aren't 😳

PR #200, opened 2021-01-31
[SE-3850] Add new template ID variable placeholders

gabor-boros:
Add new template ID variable placeholders

**Dependencies**: None

**Reviewers**
- [x] @nizarmah


nizarmah (approved):
👍🏼 

I verified that the variable names are correct, and there are no mismatches with the code.

I also deployed the changes internally.

PR #203, opened 2021-02-22
[FAL-884] Add forum and edxchange hosts

toxinu:
This pull request is about adding the OpenCraft and edxchange instances into our playbooks.

New hosts secrets: https://github.com/open-craft/ansible-secrets/pull/238
New Terraform configuration: https://github.com/open-craft/terraform-secure

**Jira tickets**

- [FAL-884](https://tasks.opencraft.com/browse/FAL-884)

**Testing instructions**

TBA

**Reviewers**

- @swalladge


playbooks/roles/discourse/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+- name: mkdir /discourse/docker
+  file:
+    path: /discourse/docker
+    state: directory
+
+- name: change docker root directory
+  lineinfile:
+    path: /lib/systemd/system/docker.service
+    regexp: '^ExecStart='
+    line: ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock -g /discourse/docker
```
swalladge, line 34: I feel like this should be a service override file, rather than changing the service inline?  eg. `/etc/systemd/system/docker.service.d/override-docker-root-dir.conf`

swalladge: Not high priority

toxinu: @swalladge You are totally right. I have changed the way to overwrite Docker settings by using `/etc/docker/daemon.json` instead.

swalladge: Ah, even better! :)

swalladge:
@toxinu :+1:

- [x] I tested this: tested results from deployments
- [x] I read through the code
- [ ] NA I checked for accessibility issues
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding
    client's `configuration-secure` repository. (in this case, ansible secrets)

swalladge (approved):

PR #205, opened 2021-03-09
[FAL-1666] Add Jira to Ansible

toxinu:
This merge request is about adding Jira provisioning to our Ansible.

Terraform: https://github.com/open-craft/terraform-secure/pull/1

**Jira tickets**

- [FAL-1666](https://tasks.opencraft.com/browse/FAL-1666)

**Testing instructions**

TBA

**Reviewers**

- @swalladge


playbooks/roles/jira/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+
+- name: install Jira dependencies
+  apt:
+    pkg:
+      - mysql-client
+
+- name: Ensure commands can be executed as www-data - set shell
+  user:
+    name: www-data
+    shell: /bin/false
```
swalladge, line 12: The name doesn't explain why the shell is set to /bin/false.?

playbooks/roles/jira/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+
+- name: Copy jira init script
```
swalladge, line 14: s/init script/service file/

playbooks/roles/jira/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  systemd:
+    name: jira
+    enabled: yes
+    daemon_reload: yes
+
+- name: add backup pruner cron job
+  cron:
+    name: clean old Jira exports
+    state: present
+    job: find /var/www/jira/var/jira/export -maxdepth 1 -type f -mtime +150 -delete -regextype sed -regex '.*/[0-9]\{4\}-[A-Z][a-z]\{2\}-[0-9]\{2\}.*.zip'
```
swalladge, line 31: If you increase the maxdepth here, you can remove the corrupted backup pruner job

playbooks/roles/node-exporter/meta/main.yml
```diff
@@ -16,3 +16,4 @@ dependencies:
     NGINX_PROXY_PUBLIC_PORT: 19100
     NGINX_PROXY_INTERNAL_PORT: 9100
     NGINX_PROXY_USE_SSL: "{{ NODE_EXPORTER_USE_SSL }}"
+    NGINX_PROXY_EXTRA_CONFIG: ""
```
swalladge, line 19: Why did this need to be added?

toxinu: To set a default value because if the `nginx-proxy` role is used at another moment in the playbook, it may use the already defined value. Which is the case for the `jira` role.

Related discussion: https://chat.opencraft.com/opencraft/pl/4hfzcctu9intxearsiy5m9akfo

playbooks/roles/jira/tasks/main.yml
toxinu: @swalladge I don't really know if Jira can create more different backups/archives with the same name pattern in this directory so I prefer to have two commands. Does it sound reasonable?

playbooks/roles/jira/tasks/main.yml
toxinu: Fixed. Thanks :)

playbooks/roles/jira/tasks/main.yml
toxinu: Task name updated. Thanks.

playbooks/roles/node-exporter/meta/main.yml
swalladge: Ah wow :grimacing: nice catch. :+1:

playbooks/roles/jira/tasks/main.yml
swalladge: But the name pattern regex you use is the same in both.

toxinu: I don't really see the problem, but anyway, I have removed this line. 👍

swalladge: There isn't really any problem I guess; it's just a lot of duplication.  Either way. :)

swalladge (approved):
@toxinu :+1: looks good to me

- [x] I tested this: this code was used to deploy production jira, so seems to be working fine
- [x] I read through the code
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding
    client's `configuration-secure` repository.

PR #206, opened 2021-03-11
Enable the Ansible profiling callback by default for OCIM installs.

janimo:
The playbooks called by OCIM will be profiled and running times for
the tasks summarized using the official profile_tasks plugin.


0x29a (approved):
:+1:

- [x] I checked that adding `ANSIBLE_CALLBACK_WHITELIST` to stage enables profiling output that can be seen in the Ocim console.

PR #207, opened 2021-03-12
Updates cert-manager playbook to use Python 3.6

nizarmah:
## Description

After the OVH server fire, we had to move the stage certificate manager to a new region.

However, when trying to run the playbook, it failed due to python 3.5.

This fixes that.

## Supporting information

* **Jira Tickets**: SE-4243

## Testing instructions

* Run the playbook on the stage certificate manager


giovannicimolin (approved):
:+1: 

This was used to deploy cert-manager stage and integration.

PR #188, opened 2020-10-14
[BB-2836] Monitoring grade report generation - Setup security features for elasticsearch and kibana

shimulch:
This PR enables security features for elasticsearch which is required to use kibana alert system.
Following new variables are introduced, which I think should be set from configuration secure repository -

- ``elasticsearch_password`` - Password that will be used for ``elastic`` and ``kibana_system`` user in elastic.
- ``kibana_elasticsearch_password`` - Password for ``kibana_system`` user to connect with elasticsearch. Essentially same as ``elasticsearch_password``.
- ``kibana_encryption_key`` - Kibana uses for encrypting and decrypting sensitive information.
- ``elasticsearch_pem_file`` - Elastic search CA pem file location for kibana.

This PR also introduces a new role - ``kibana-email-alert``. Which enables a ``systemd`` service that watches over Kibana logs on ``journald`` with a predefined regex. If there a Kibana Alert with ``Server Log`` connector and that Alert triggers, this new service collects that log and notifies to a given email address. New variables for this role -

- ``kibana_email_alert_regex`` - what to match in journald log, defaults to ``Server log``
- ``kibana_email_alert_to``- which email to send alert, **Required**
- ``kibana_email_alert_subject`` - email subject, defaults to "Kibana Error Log"

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-2836

~~**Discussions**:~~

**Dependencies**: None

~~**Screenshots**:~~

~~**Sandbox URL**:~~

**Merge deadline**: "None" 

**Testing instructions**:

1. Deploying Kibana should now setup elasticsearch bootstrap password and ``kibana_system`` user password
2. Kibana should be able to connect to elasticsearch using that password
3. Alerting feature is enabled in Kibana
4. Create a ``Server Log`` connector
5. Test that connector if email alert working
6. Create a Kibana Alert & set the ``Server Log`` connector as action
7. When Alert triggers, an email should go to the address defined in ``kibana_email_alert_to`` variable

**Author notes and concerns**:
1. Email contents are not properly formatted
2. Is there a better regex other than ``Server log``?

**Reviewers**
- [ ] @kaizoku 

~~**Settings**~~

PR #208, opened 2021-03-15
FAL-718 fix elasticsearch deployment

swalladge:
[FAL-718](https://tasks.opencraft.com/browse/FAL-718)

- fix backward compatibility with existing elasticsearch setups (conditionally enable the new password/keystore/kibana things)
- support elasticsearch 7 (via extra cfg; es 7 requires `discovery.type` set)
- add standard .tool-versions file with a supported python version (so i stop accidentally using python 3.9 which isn't supported here yet)

**Test instructions**:

This was used to deploy the elasticsearch-labxchange-ci instance (see https://github.com/open-craft/ansible-secrets/pull/246 ), so just need to verify that PR.

Also read through and sanity check all the changes - it needs to work with the new setup from https://github.com/open-craft/ansible-playbooks/pull/188 , as well as the original setup for elasticsearch 1.5 and dedicated es/memcached instances for ocim instances.  Ideally, we would run the playbooks against one of the new ES servers with the kibana setup, and one of the older ES servers used with ocim instances, but not sure if that's a good idea - if there's a bug, it will break the production servers. :grimacing: 

**Reviewers**:

- [ ] @toxinu


swalladge:
@shimulch could you add your input here please?  I'm having troubles getting things working. :)

playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -28,27 +28,31 @@
     - "{{ elasticsearch_log_dir }}"
     - "{{ elasticsearch_cfg_dir }}"
 
-- name: Install Elasticsearch certificates
-  copy:
-    content: "{{ elasticsearch_keystore }}"
+- name: Install Elasticsearch certificates 1
+  template:
+    src: "http.p12.b64.j2"
```
swalladge, line 33: `copy` fails if content is empty, and the docs specifically say not to use this for a single variable:

> If you need variable interpolation in copied files, use the ansible.builtin.template module. Using a variable in the content field will result in unpredictable output.

toxinu:
@swalladge :+1:

Thanks for adding ES 7 support \o/

- [ ] I tested this: nothing
- [x] I read through the code
- [ ] I tested that the UI can be used with a keyboard only (tab order, keyboard controls).
- [ ] Includes documentation

PR #209, opened 2021-03-19
Add default node-exporter values so it doesn't get overwritten

toxinu:
This merge request is about fixing an unwanted variable overwritting when using `nginx-proxy` along with `common-server` like for the Jira and Forums playbooks: https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/forum.yml

If [this variable](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/forum.yml#L17) is being overwritten in `nginx-proxy` role, the same value will be used while deploying `node-exporter` role from `common-server`.

@itsjeyd This is related to the issue you reported to be about forums and jira node-exporters being unavailable. May you review this ticket just by checking that Prometheus is now happy with the fix?

**Jira tickets**

- [FAL-884](https://tasks.opencraft.com/browse/FAL-884)


itsjeyd (approved):
@toxinu Thanks for getting this fix out so quickly 🚀 The PR LGTM :+1:

- [x] I tested this: Works as advertised in PR description. (Verified that Prometheus reported `instance_down` alert as resolved, and double-checked status of affected servers on targets page.)
- [x] I read through the code

PR #212, opened 2021-03-29
[FAL-871] Fixes for upgrading mongodb to 4.0

pomegranited:
* Disables the cloud monitoring which is [enabled by default in mongodb>4.0](https://docs.mongodb.com/manual/administration/free-monitoring/).
   If we want to use this, we need to evaluate security and determine how to configure it.
* Upgrades the mongodb_exporter package installed for use with Prometheus.
   Older versions don't work anymore due to changes in the go dependencies.

**Testing instructions**

See https://github.com/open-craft/ansible-secrets/pull/255

**Reviewer**

- [x] @giovannicimolin


playbooks/roles/mongodb/defaults/main.yml
```diff
@@ -8,7 +8,7 @@ MONGODB_SERVER_BACKUP_DIR: /var/lib/mongodb/backup
 # PROMETHEUS ##################################################################
 
 MONGODB_EXPORTER_ENABLED: true
-MONGODB_EXPORTER_VERSION: "v0.6.2"
+MONGODB_EXPORTER_VERSION: "v0.20.1"
```
giovannicimolin, line 11: @pomegranited I've checked stage and the mongodb exporter isn't working. Probably because of the huge leap in the version here.

Upon reading the docs and testing a few things, I found that we need to make a few changes to make it work again.
I've got `mongodb-exporter` up again with the following parameters:
```
# Notice the double `-` on the parameters.
# `compatible-mode` flag is required so that the exporter 
# logs the same information as the old version and gives us 
# some time before we have to fix our grafana dashboards. 
/usr/local/bin/mongodb_exporter \
  --web.listen-address=":9001" \
  --compatible-mode \
  --mongodb.uri=mongodb://<USERNAME>:<PASSWORD>@localhost:27017/admin
```

I've updated the service file manually on the server to get it back up.

giovannicimolin (changes requested):
Found a minor issue with MongoDB exporter.

playbooks/roles/mongodb/defaults/main.yml
pomegranited: Thank you for fixing this @giovannicimolin ! If I could have gotten an earlier version to build, I would have done that, but this was the only one whose `go` dependencies worked.

I've updated the template to use the settings you provided, and redeployed to stage to test.

giovannicimolin (approved):
:+1: 

- [x] I tested this: checked that stage mongodb is on version 4.0 and that the mongodb-exporter is working.
- [x] I read through the code

PR #215, opened 2021-03-30
[FAL-1667] Pin jinja2 to a version that works with filebeat

tartansandal:
The `filebeat` role requires a minimum version of `jinja2` in order to complete, due to the particular invocation of the `indent` filter.  Only tested this on the newer Ansible though.

Test instructions:

1.


eLRuLL:
@tartansandal approved :+1:

PR #210, opened 2021-03-22
[BB-3847] Update prometheus role to support LimitNOFILE

shimulch:
## Description

This PR updates the `ernestas-poskus.prometheus` role from the submodule repository https://github.com/ernestas-poskus/ansible-prometheus.

It fixes `too many open files error` in prometheus.net.opencraft.hosting. A [PR](https://github.com/ernestas-poskus/ansible-prometheus/pull/105) was created on the third-party repo to solve the issue, which got merged. This PR is to update the corresponding submodule to include the fix.

## Supporting information

https://tasks.opencraft.com/browse/BB-3847

## Testing instructions

1. Pull this PR
2. Deploy a prometheus instance
3. The `prometheus.service` file should contain `LimitNOFILE` set to `infinity`

## Deadline

"None" 

## Other information

**Reviewers**
- [ ] @0x29a

PR #221, opened 2021-05-04
Update prometheus role to increase nofile limit for service

xitij2000:
The default nofile limit for prometheus is quite low, and the upstream change with the fix comes with a lot of other changes that we can't apply right now.

This will override the prometheus systemd unit to apply this limit without editing the existing file.

### DIFF
--- a/playbooks/roles/prometheus/tasks/main.yml
+++ b/playbooks/roles/prometheus/tasks/main.yml
@@ -16,4 +16,11 @@
     src: logrotate
     dest: /etc/logrotate.d/prometheus
 
+- name: set the nofile limit for prometheus service to a high number
+  copy:
+    content: |
+      [Service]
+      LimitNOFILE=infinity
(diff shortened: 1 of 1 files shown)


shimulch:
@xitij2000 I think this is related to https://github.com/open-craft/ansible-playbooks/pull/210

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Added a tag to the extra task.
2. Ran the playbooks with the specific tag.
3. Checked that the configuration was deployed to the prometheus server.
- [x] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation

@xitij2000 LGTM.

xitij2000:
> @xitij2000 I think this is related to #210

It does, but my understanding is that PR can't merge for quite some time till we upgrade other parts of our infrastructure. This can solve the issue for now, and we can remove this patch when the other PR merges.

PR #222, opened 2021-05-07
FAL-1750 Add playbook for mailing list host

swalladge:
[FAL-1750](https://tasks.opencraft.com/browse/FAL-1750)

This adds a playbook for mailing list tasks.  We'll need to add to this later when we set up mailman 3, etc.

**Test instructions**:

See https://github.com/open-craft/ansible-secrets/pull/267

Basically, this should set up the common server role on the new mailing list server.

**Reviewers**:

- [x] @toxinu


toxinu:
Pretty basic, good to merge :+1:

PR #223, opened 2021-05-31
FAL-1751 Configure mailing list software

swalladge:
[FAL-1751](https://tasks.opencraft.com/browse/FAL-1751)

Add ansible scripts for configuring a mailing list.

Work in progress.

**Test instructions**:

- Set configuration variables
- Run this against an instance
- Verify that mailman3, mailman3-web, nginx proxy, postfix are all set up.

**Reviewers**:

- [x] @toxinu

### DIFF
--- a/playbooks/roles/mailing-list/handlers/main.yml
+++ b/playbooks/roles/mailing-list/handlers/main.yml
@@ -1,0 +1,25 @@
+---
+- name: "restart mailman3"
+  service:
+    name: "{{ item }}"
+    state: restarted
+  with_items:
+    - "mailman3"
+    - "mailman3-web"
+
+- name: "reload postfix"
+  service:
+    name: postfix
+    state: reloaded
+
+# https://docs.mailman3.org/projects/mailman/en/latest/src/mailman/docs/mta.html#transport-maps
+# > Mailman will automatically write the correct transport map when its mailman aliases command is run,
+# > or whenever a mailing list is created or removed via other commands.
+#
+# So we have this here as a handler to be called when mailman is installed, updated, config changed, etc. from ansible.
+# In normal operation, mailman will keep it up to date automatically.
+#
+# This generates the maps databases used by postfix in /var/lib/mailman3/data/
+- name: "update mailman3 aliases"
+  command: "mailman aliases"
+  become_user: "list"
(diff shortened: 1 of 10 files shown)


playbooks/roles/mailing-list/templates/postfix-main.cf
```diff
@@ -0,0 +1,29 @@
...
+# Only relay mail from the local mailman server
+mynetworks_style = host
+
+# Support the default VERP delimiter.
+recipient_delimiter = +
+
+# Don't treat owner- and -request addresses specially
+owner_request_special = no
+
+# Let mailman3 manage the transport and relay
```
swalladge, line 23: @tartansandal I thought postfix was supposed to do the transport part?

tartansandal: >  thought postfix was supposed to do the transport part?

@swalladge Yup that is exactly what we are doing.  What the above setting does is allow `mailman` to manage how `postfix` handles the transport (by changing its configuration). This way you can add/modify mailing lists in mailman without having to explicitly touch the postfix transport configuration. Make sense?

playbooks/roles/mailing-list/handlers/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: "restart mailman3"
+  service:
+    name: "{{ item }}"
+    state: restarted
```
tartansandal, line 5: This should really be two separate handlers.

playbooks/roles/mailing-list/handlers/main.yml
```diff
@@ -0,0 +1,13 @@
...
+  with_items:
+    - "mailman3"
+    - "mailman3-web"
```
tartansandal, line 8: @swalladge This should really be two separate handlers (`notify` takes a list). And you don't need all the quotes.

playbooks/roles/mailing-list/templates/postfix-main.cf
swalladge: Yep, cool :+1:

toxinu (approved):
:+1:

- [x] I tested this: just browse the new playbook and role
- [x] I read through the code

PR #227, opened 2021-06-12
[FAL-1995]: adds partition check for mounted /var

eLRuLL:
This adds special partitions (like the case in the MySQL Servers hosted in SoYouStart with the `mysql` partition name, but mounted for `/var/lib/mysql`) to be checked by the sanity checks

### DIFF
--- a/playbooks/roles/sanity-checker/files/sanity_check.py
+++ b/playbooks/roles/sanity-checker/files/sanity_check.py
@@ -98,7 +98,10 @@ def get_dev_mounts():
     for line in proc_mounts.split("\n"):
         parts = line.split()
         mount = Mount(parts[1], parts[0])
-        if mount.device.startswith("/dev"):
+        if (
+                mount.device.startswith("/dev")
+                or mount.mount.startswith("/var/")      # special cases, but also used by the system
+        ):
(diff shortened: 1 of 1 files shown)


playbooks/roles/sanity-checker/files/sanity_check.py
```diff
@@ -98,7 +98,10 @@ def get_dev_mounts():
     for line in proc_mounts.split("\n"):
         parts = line.split()
         mount = Mount(parts[1], parts[0])
-        if mount.device.startswith("/dev"):
+        if (
+                mount.device.startswith("/dev")
+                or mount.mount.startswith("/var/")      # special cases, but also used by the system
```
swalladge, line 103: @eLRuLL could you update the docstring for this method to note this case, and also add a comment about why the /var/ mount device name doesn't start with /dev?

eLRuLL: @swalladge I think the docstring is still correct, this is not for a "device" with that name, it's a check to confirm it's mounted on system folder. I added more explanation in the comment.

swalladge (approved):
@eLRuLL :+1:

- [x] I tested this: checked on mysql slave 1
- [x] I read through the code
- [x] Includes documentation

PR #228, opened 2021-06-16
FAL-1752 more setting up for mail server and mailing list

swalladge:
[FAL-1752](https://tasks.opencraft.com/browse/FAL-1752)

Add/update configuration for a mail server that hosts a mailing list (mailman3 + mailman3 web tools), and postfix config for aliases and forwarding mail.  It's relatively simple because this is for our particular use case; this isn't a general 'mail server' or 'mailing list server' role.

**Test instructions**:

- verify that the defaults are sane and documented
- read through the playbook and verify that all the steps make sense, the correct handlers are called, etc.
- check the postfix config

**Reviewers**:

- [x] @toxinu


playbooks/roles/sanity-checker/files/sanity_check.py
```diff
@@ -86,28 +87,36 @@ def send_message(recipient, subject, body):
 
 def get_dev_mounts():
     """
-    :return: List of mounted drives, we assume that drives will have
```
toxinu, line 89: Are these changes related to the pull request?

toxinu (approved):
I just left one comment about a change but it looks good to merge :+1:

- [x] I tested this: checked the roles and the postfix configuration
- [x] I read through the code
- [x] Includes documentation

playbooks/roles/sanity-checker/files/sanity_check.py
swalladge: @toxinu yes, please see https://github.com/open-craft/ansible-playbooks/pull/228/commits/0aaf9faf1553e3fee9f8cf8304c10fdb7031661f

It was previously broken on mail.opencraft.com because it was trying to get fs info on /var/lib/lxcfs

PR #231, opened 2021-07-02
BB-4426: Replace networking with netplan configuration WIP

giovannicimolin:
This PR replaces `networking` for `netplan` configuration and makes this playbook compatible with Ubuntu 20.04.
I don't want to keep multiple release version handling for simplicity, so this should only be merged once we have the bare-metal servers on Focal.

**Testing instructions:**
I've already applied this playbook on our spare baremetal server.

1. SSH into the server using the DNS record, IP, and failover IP address.
2. Check the network configuration with `ifconfig`.

**Notes:**
Since there's some secrets involved in testing this, I'll leave a comment in the internal ticket.

**Reviewers:**
- [ ] @0x29a 
- [x] @lgp171188

### DIFF
--- a/playbooks/roles/common-server-init/tasks/networking.yml
+++ b/playbooks/roles/common-server-init/tasks/networking.yml
@@ -8,8 +8,6 @@
       template:
         src: network/interfaces.d/floating-ip.cfg
         dest: /etc/network/interfaces.d/floating-ip.cfg
-        src: netplan/floating-ip.yaml
-        dest: /etc/netplan/floating-ip.yaml
         owner: root
         group: root
         mode: 0644
--- a/playbooks/roles/common-server-init/tasks/networking.yml
+++ b/playbooks/roles/common-server-init/tasks/networking.yml
@@ -1,13 +1,15 @@
 ---
 
-- name: Add network interface for floating IP
-  when: COMMON_FLOATING_INTERFACE_NAME and COMMON_FLOATING_IP
-  block:
+- name: Add network interface for floating IP (Xenial, Bionic)
+  when: COMMON_FLOATING_INTERFACE_NAME and COMMON_FLOATING_IP and ansible_distribution_release in ["xenial", "bionic"]
 
+  block:
     - name: Add floating IP as network interface alias
       template:
         src: network/interfaces.d/floating-ip.cfg
         dest: /etc/network/interfaces.d/floating-ip.cfg
+        src: netplan/floating-ip.yaml
+        dest: /etc/netplan/floating-ip.yaml
         owner: root
         group: root
         mode: 0644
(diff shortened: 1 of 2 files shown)


playbooks/roles/common-server-init/templates/netplan/floating-ip.yaml
```diff
@@ -0,0 +1,12 @@
+network:
+  version: 2
+  renderer: networkd
+  ethernets:
+    {{ COMMON_FLOATING_INTERFACE_NAME }}:
+      dhcp4: yes
+      dhcp6: yes
```
lgp171188, line 7: @giovannicimolin, with DHCP enabled here, and a static IP set in the `addresses` list below, does netplan add the IP address received from DHCP (which is also a static IP) to the same interface without overwriting or affecting the floating IP address?

giovannicimolin: Yes. Netplan accepts this kind of setup even without using the `label` attributes that I've put.

lgp171188 (commented):
@giovannicimolin, apart from a single question, these changes look okay to me. ~~Is there a physical server with a floating IP where this has been tested and verified to work okay?~~ Nvm, found out about this from the task :) CC @0x29a

playbooks/roles/common-server-init/tasks/networking.yml
```diff
@@ -1,13 +1,15 @@
 ---
 
-- name: Add network interface for floating IP
-  when: COMMON_FLOATING_INTERFACE_NAME and COMMON_FLOATING_IP
-  block:
+- name: Add network interface for floating IP (Xenial, Bionic)
+  when: COMMON_FLOATING_INTERFACE_NAME and COMMON_FLOATING_IP and ansible_distribution_release in ["xenial", "bionic"]
```
lgp171188, line 4: @giovannicimolin, do we have any dedicated servers running Bionic? If not, we should drop it from the check.

playbooks/roles/common-server-init/tasks/networking.yml
```diff
@@ -1,13 +1,15 @@
...
 
+  block:
     - name: Add floating IP as network interface alias
       template:
         src: network/interfaces.d/floating-ip.cfg
         dest: /etc/network/interfaces.d/floating-ip.cfg
+        src: netplan/floating-ip.yaml
+        dest: /etc/netplan/floating-ip.yaml
```
lgp171188, line 12: @giovannicimolin, why is this needed on Xenial?

lgp171188 (approved):
:+1: 
- [x] I checked these changes by checking the spare server used for testing this PR and 
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ NA
- [x] ~~Includes documentation~~ NA

playbooks/roles/common-server-init/tasks/networking.yml
giovannicimolin: @lgp171188 I think we have one or two servers that were upgraded to Bionic due to dependency issues. I'll keep it around just in case. 

When the upgrades are done, we should clear this entire block and remove the check for the Focal branch.

playbooks/roles/common-server-init/tasks/networking.yml
giovannicimolin: @lgp171188 Oops. This is a typo, fixed!

PR #235, opened 2021-07-15
Fix copy-pasted description of prometheus alert.

mtyaka:
This fixes the error message of `haproxy_frontend_request_errors_high` which seems to have been copy-pasted from `haproxy_backend_response_time_high` and never properly updated.

**Testing instructions**:

1. Log into https://prometheus.net.opencraft.hosting/alerts
2. Click on `haproxy_frontend_request_errors_high` and verify that the summary and description templates looks correct.


toxinu:
@mtyaka Looks good to merge 👍

PR #237, opened 2021-07-22
FAL-2085 Increase tmpreaper runtime limit by default

swalladge:
In the process, refactor tmpreaper config tasks to use a template,
because we can no longer neatly use the lineinfile method.

**Test instructions**:

- run this playbook against ocim stage and verify the new task shows as unchanged (this has already been run on ocim stage) - note that ocim stage is now running focal, so you'll need the ansible 2.8 requirements set.
- ssh into ocim stage and check that /etc/tmpreaper.conf is as expected and has the expected runtime limit override
- verify that we aren't getting pinged on ops about `error: run time exceeded!` for tmpreaper on ocim stage

**Reviewers**:

- [ ] @mtyaka

### DIFF
--- a/playbooks/roles/common-server-init/defaults/main.yml
+++ b/playbooks/roles/common-server-init/defaults/main.yml
@@ -1,1 +1,5 @@
 COMMON_SERVER_INSTALL_NGINX: true
+
+# Default tmpreaper runtime limit is 55 seconds.
+# This often isn't enough, so increase by default.
+TMPREAPER_RUNTIME: '110'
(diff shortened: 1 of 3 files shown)


mtyaka:
Nice work @swalladge and @alfredchavez :+1:

- [x] I tested this: I verified that the changes have been applied to Ocim stage and prod. I verified that we are no longer receiving error notifications about exceeded runtime to ops@ mailing list. Note: I didn't actually run the playbook myself because I don't currently have access to my main computer and don't want to spend too much time setting up ansible playbooks on my temporary laptop. The updated code has already been run successfully on two VMs, so I don't think it's necessary for me to re-run it.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ n/a
- [x] Includes documentation (inline comments)

PR #238, opened 2021-07-22
FAL-2116 Add support for configuring postfix header_checks

swalladge:
https://tasks.opencraft.com/browse/FAL-2116

As in title.

**Test instructions**:

- run the playbook
- verify that /etc/postfix/header_checks is templated out as expected

**Reviewers**:

- [ ] TODO


swalladge:
Closing because we didn't end up needing this.

PR #240, opened 2021-08-07
[SE-4608] set apt preferences to  prioritize the opencraft repository

hchandad:
This change prioritizes using the opencraft repository for the chkrootkit package. This way future updates to the main release wouldn't prioritize the main release over the opencraft one.

### DIFF
--- a/playbooks/roles/common-server-init/tasks/misc.yml
+++ b/playbooks/roles/common-server-init/tasks/misc.yml
@@ -28,6 +28,14 @@
     id: "{{ COMMON_SERVER_OPENCRAFT_REPO_KEY_ID }}"
     state: present
 
+- name: set apt preferences to prioritize the opencraft repo
+  copy:
+    content: |
+            Package: chkrootkit
+            Pin: Origin "debs.opencraft.com"
+            Pin-Priority: 1001
+    dest: /etc/apt/preferences.d/opencraft
+
 - name: apt-get update, dist-upgrade and autoremove
   apt:
     update_cache: yes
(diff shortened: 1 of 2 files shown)


playbooks/roles/common-server-init/tasks/misc.yml
```diff
@@ -17,17 +17,25 @@
...
+- name: add Opencraft repository
+  apt_repository:
+    repo: "{{ COMMON_SERVER_OPENCRAFT_REPO }}"
+    state: present
+
+- name: set apt preferences to prioritize the opencraft repo
+  copy:
+    content: |
+            Package: chkrootkit
+            Pin: Origin "debs.opencraft.com"
```
arbrandes, line 35: @hchandad, can you please factor out debs.opencraft.com into its own variable (like COMMON_SERVER_OPENCRAFT_REPO_DOMAIN) and use it here?  This way we avoid problems if by any chance the domain changes, some day.

hchandad: Done !

playbooks/roles/common-server/defaults/main.yml
```diff
@@ -52,6 +52,7 @@ COMMON_SERVER_NO_BACKUPS: false
 COMMON_SERVER_OPENCRAFT_REPO: "deb http://debs.opencraft.com/ {{ ansible_distribution_release }} main"
```
arbrandes, line 52: Sorry I wasn't more explicit, but the idea was to use COMMON_SERVER_OPENCRAFT_REPO_DOMAIN here, too.

hchandad: It has been updated as well !

arbrandes (approved):
LGTM, @hchandad!

PR #242, opened 2021-08-20
FAL-2311 Run the sanity checker every 10 minutes

swalladge:
[FAL-2311](https://tasks.opencraft.com/browse/FAL-2311)

Previously, it was _supposed_ to run every 10 minutes,
but it skipped '0', so there was a 20min gap every hour,
which can cause false positives for snitches.

Also don't use the `*/10` syntax,
because we want to offset it by one minute,
to avoid issues where everything happens on a minute divisible by 10.

**Test instructions**:

- run this on a server that has the sanity checker configured
- wait an hour
- verify that it runs every 10 minutes, starting at 1 minute past the hour

**Reviewers**:

- [x] @clemente

### DIFF
--- a/playbooks/roles/sanity-checker/tasks/main.yml
+++ b/playbooks/roles/sanity-checker/tasks/main.yml
@@ -18,5 +18,5 @@
 - name: install sanity checks crontab
   cron:
     job: "{{ SANITY_CHECK_SCRIPT_LOCATION }} {{ SANITY_CHECK_CONFIG }} > /dev/null 2>&1"
-    minute: "10,20,30,40,50"
+    minute: "1,11,21,31,41,51"
     name: "sanity checker for: {{ SANITY_CHECK_SUBJECT }}"


clemente:
A bit of research about the status before the PR (when we were missing the :00 ping) :
- DeadManSnitch works in 15-minute periods like 14:30--14:45,  14:45--15:00,  15:00--15:15.
- a 20 minute hiatus isn't as bad as it seems as long as it happens on the right times. For instance, if the cron job takes 1 minute and the pings are sent at :41, :51, [not at :01], :11, :21, … then everything is fine because each DMS window received a ping. In particular the 00-15 window got a ping at :11
- However, if just one of the cron jobs (the :11 one) takes a few more minutes for some reason (let's say 5 more minutes), then we get :41, :51, [not at :01], :16, :21, … and there's a missing ping in the 00-15 window, and we get an alert
- this means that the current system (which lacked the :00 ping) is fragile because a single 5 or 6 minutes delay could make it page
- and I'm not only referring to a 5 or 6 minute delay. If the computer clock is off by 5 or 6 minutes, the same thing happens  

> Also don't use the */10 syntax,
> because we want to offset it by one minute,
> to avoid issues where everything happens on a minute divisible by 10.

I think this is important, for another reason: if a ping to DMS runs exactly at 15:00, it's not clear which window will it fall into (14:45--15:00, vs. 15:00-15:15); it may depend on how well set is each clock and whether DMS thinks it's 14:59:59 or 15:00:00 or 15:00:01. By using 01,11,21,31,41,51 it's clearer.

For the above reasons: yes, let's fix this, even if the issues may not be very common.

@swalladge 

:+1: I read the code. Documentation not needed there, but the commit history will refer to the task. No other issues with the approach; it feels safer with „every 10 minutes“.

By the way, [in the snitch](https://deadmanssnitch.com/snitches/d3f574912a) we don't see the new :01 ping in the 00-15 window because it's superseded by the :11 on (DMS seems to show the latest ping in each window).

I'll continue about how to deploy this, in the task.

PR #244, opened 2021-09-09
[BB-4567] fix: pass the apt packages list directly to the apt module

lgp171188:
The previous implementation used the loop keyword which installed the
provided packages one at a time and hence was inefficient.


farhaanbukhsh (approved):
👍 

- ✅  I tested this, this is already deployed
- ✅  I read through the code
- ❌  Includes documentation

PR #247, opened 2021-09-24
fix: increase nginx max_body_size for mailman [BB-4955]

Agrendalath:
The default size (1M) was causing lots of 413 errors.

I thought about making this configurable, but we're not resuing this role, so we can keep 10M as a default value for now.


farhaanbukhsh (approved):
👍 

- ✅  I tested this: we fixed this on the server
- ✅  I read through the code
- ❌  Includes documentation

PR #248, opened 2021-10-06
[FAL-2354] Increase PROMETHEUS_ALERT_HTTP_RESPONSE_FRONTEND_ERROR_THRESHOLD to 50

mtyaka:
35 is was too low and resulted in too many false positives.

**Testing instructions:**

- Look at the `haproxy_frontend_request_errors_high` on the [prometheus alerts page](https://prometheus.net.opencraft.hosting/alerts).
- Explore the [results of the alert query](https://prometheus.net.opencraft.hosting/graph?g0.range_input=2w&g0.expr=(100%20*%20(rate(haproxy_frontend_request_errors_total%7Bfrontend%3D%22fe_https%22%7D%5B5m%5D)%20%2F%20rate(haproxy_frontend_http_requests_total%7Bfrontend%3D%22fe_https%22%7D%5B5m%5D)))&g0.tab=0) and check whether the new limit of `50` makes sense.


mtyaka:
Thanks for checking @toxinu :)

PR #256, opened 2021-11-17
[SE-4969] Hx Annotation tools upgrade

gabor-boros:
## Description

This PR replaces the playbooks and roles for the Hx Annotation toolset.

## Supporting information

As part of this change the main goal is to upgrade HxAT and CatchPy. Keeping Loris is not needed, hence we can delete its playbook. Also, we are aiming to reuse as much of our existing playbooks as we can, therefore we need to upgrade the `geerlingguy.postgresql` sumbodule and extend our `postgresql` roles to meet the expectations.

About the geerlingguy.[mysql, postgresql] updates, there were multiple conversations:

* https://chat.opencraft.com/opencraft/pl/ya9x7cjy13d3zfbm7gm8rqeqdh
* https://tasks.opencraft.com/browse/BB-4048

_In the case of MySQL, we only update the remote, not the playbooks._

### Dependencies

N/A

## Testing instructions

1. Read the code

## Deadline

ASAP


joidegn:
LGTM :+1: 

- [x] Read the code
- [ ] ~~Did not test it~~

pkulkark (approved):
LGTM :+1: 

- [x] I tested this: verified that the service deployed using this works as expected
- [x] I read through the code

shimulch:
@gabor-boros, in this PR you've deleted `hxat-catchpy-loris.yml` playbook. But it still referred to [deploy-all.yml](https://github.com/open-craft/ansible-playbooks/blob/master/playbooks/deploy-all.yml#L9). Due to this it throws error while running `deploy-all` playbook.

PR #257, opened 2021-11-18
Update mailman configuration to match latest version of mailman on 20.04 [BB-4467]

xitij2000:
Update mailman3 configuration to work with the latest release of mailman3 on Ubuntu 20.04.

### DIFF
--- a/playbooks/roles/mail/templates/mailman-web.py.j2
+++ b/playbooks/roles/mail/templates/mailman-web.py.j2
@@ -32,7 +32,6 @@ INSTALLED_APPS = (
     'django.contrib.staticfiles',
     'rest_framework',
     'django_gravatar',
-    'paintstore',
     'compressor',
     'haystack',
     'django_extensions',
@@ -72,3 +71,4 @@ USE_TZ = True
 
 DEFAULT_FROM_EMAIL = '{{ MAILMAN_FROM_EMAIL }}'
 SERVER_EMAIL = '{{ MAILMAN_FROM_EMAIL }}'
+POSTORIUS_TEMPLATE_BASE_URL = 'https://{{ MAILMAN_WEB_DOMAIN }}'
--- a/playbooks/roles/mail/templates/mailman3-nginx-proxy.conf.j2
+++ b/playbooks/roles/mail/templates/mailman3-nginx-proxy.conf.j2
@@ -1,7 +1,7 @@
 # {{ ansible_managed }}
 
 upstream mailman3 {
-    server unix:/run/mailman3/web/uwsgi.sock fail_timeout=0;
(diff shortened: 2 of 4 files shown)


playbooks/roles/mail/templates/mailman-web.py.j2
```diff
@@ -32,7 +32,6 @@ INSTALLED_APPS = (
     'django.contrib.staticfiles',
     'rest_framework',
     'django_gravatar',
-    'paintstore',
```
xitij2000, line 35: This is no longer used with the newer version of mailmain3

playbooks/roles/mail/templates/mailman-web.py.j2
```diff
@@ -72,3 +71,4 @@ USE_TZ = True
 
 DEFAULT_FROM_EMAIL = '{{ MAILMAN_FROM_EMAIL }}'
 SERVER_EMAIL = '{{ MAILMAN_FROM_EMAIL }}'
+POSTORIUS_TEMPLATE_BASE_URL = 'https://{{ MAILMAN_WEB_DOMAIN }}'
```
xitij2000, line 74: This is a new configuration value.

playbooks/roles/mail/templates/mailman3-nginx-proxy.conf.j2
```diff
@@ -1,7 +1,7 @@
 # {{ ansible_managed }}
 
 upstream mailman3 {
-    server unix:/run/mailman3/web/uwsgi.sock fail_timeout=0;
+    server unix:/run/mailman3-web/uwsgi.sock fail_timeout=0;
```
xitij2000, line 4: The location of this socket has changed in either ubunutu's config or the upsteream itself.

swalladge:
@xitij2000 while you're here, could you also remove the [block device mount](https://github.com/open-craft/ansible-playbooks/blob/a2ec8d9fd1bce40db44f05fdd8b7505d71f19a48/playbooks/roles/mail/tasks/main.yml#L3-L9)?  We no longer use it, since the latest redeployment.  CC @giovannicimolin

xitij2000:
> @xitij2000 while you're here, could you also remove the [block device mount](https://github.com/open-craft/ansible-playbooks/blob/a2ec8d9fd1bce40db44f05fdd8b7505d71f19a48/playbooks/roles/mail/tasks/main.yml#L3-L9)? We no longer use it, since the latest redeployment. CC @giovannicimolin

Done.

0x29a (approved):
:+1:

- [x] I tested that mail server is working fine after the upgrade.
- [x] I read through the diff.

PR #259, opened 2021-11-22
fix: geerlingguy.mysql role commit hash

gabor-boros:
## Description

This PR fixes the MySQL playbook GitHub URL. Upstream does not contain the commit hash required by Ocim.

```
  cmdline: git reset --hard 57c8638cc5e0659165c5e6b8c2f0f60884e29b46 --
  stderr: 'fatal: Could not parse object '57c8638cc5e0659165c5e6b8c2f0f60884e29b46'.'

The above exception was the direct cause of the following exception:

  File "/var/www/opencraft/instance/models/openedx_appserver.py", line 640, in provision
    return self._do_provision()
  File "/var/www/opencraft/instance/models/openedx_appserver.py", line 687, in _do_provision
    raise ProvisioningError("AppServer deploy failed: unhandled exception") from ex
instance.models.openedx_appserver.ProvisioningError: AppServer deploy failed: unhandled exception
```


toxinu:
Did they rewrite the git history so that hashes changed?
Anyway, good to merge 👍

gabor-boros:
@toxinu No, they did not. It was me trying to update the upstream of the repo and messed it up. After a revert I forgot to update the commit hash and here we go.

Thank you for the approve!

PR #262, opened 2022-01-27
[SE-5257] update external dependency

karimelghazouly:
issue: [SE-5257](https://tasks.opencraft.com/browse/SE-5257)
reference: [new linux break node_exporter `diskstats` parsing](https://github.com/prometheus/node_exporter/issues/1049)

We haven't updated this sub module for 4 years, right now we have in our server a very old version of node_exporter(0.15.2) while the most recent one is (1.3.1).
Since we use ubuntu 20 which uses linux 4.19+ so we need to update node_exporter to support this.
we install node_exporter from an external repo added as a sub module so i just updated this sub module in thi PR


gabor-boros:
👍  🎉  Great PR Karim, thank you!

- [x] I tested the role updates
- [x] I read through the code

PR #263, opened 2022-02-11
[SE-5324] new alert rules for updated node exporter

karimelghazouly:
Description
This PR add new alert rules for updated node exporter as the new node exporter versions use slightly different metric names

Supporting information
[ticket|https://tasks.opencraft.com/browse/SE-5324]

Dependencies
N/A

Screenshots
N/A

Testing instructions
I already manually add new alert rules config file on prometheus server you can find the new alerts in the prometheus alerts tab, I tested that each alert fire successfully on an appserver that has updated node exporter

Deadline
N/A

### DIFF
--- a/playbooks/group_vars/prometheus/public.yml
+++ b/playbooks/group_vars/prometheus/public.yml
@@ -91,7 +91,7 @@ prometheus_rules:
   - name: NodeRules_v2
     # We created this to support alerts for nodes that have node exporter version > 0.15 
     rules:
-    - alert: "cpu_usage_high"
+    - alert: "cpu_usage_high_v2"
       expr: "(100 * (1 - (rate(node_cpu_seconds_total{mode='idle'}[5m])))) > {{ PROMETHEUS_ALERT_CPU_THRESHOLD }}"
       for: "5m"
       labels:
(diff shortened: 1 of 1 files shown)


playbooks/group_vars/prometheus/public.yml
```diff
@@ -88,6 +88,46 @@ prometheus_rules:
       annotations:
         summary: "Instance [[ $labels.node ]] ([[ $labels.instance ]]) disk is filling up."
         description: "Instance disk may fill up {{ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD }}% in 4 hours."
+  - name: NodeRules_v2
+    # We created this to support alerts for nodes that has node exporter version > 0.15
```
mtyaka, line 92:
```suggestion
    # We created this to support alerts for nodes that have node exporter version > 0.15 
```

playbooks/group_vars/prometheus/public.yml
```diff
@@ -88,6 +88,46 @@ prometheus_rules:
       annotations:
         summary: "Instance [[ $labels.node ]] ([[ $labels.instance ]]) disk is filling up."
         description: "Instance disk may fill up {{ PROMETHEUS_ALERT_DISK_SPACE_THRESHOLD }}% in 4 hours."
+  - name: NodeRules_v2
+    # We created this to support alerts for nodes that has node exporter version > 0.15 
+    rules:
+    - alert: "cpu_usage_high"
```
mtyaka, line 94: @karimelghazouly Alerts that you manually defined on the prometheus server have a `_v2` suffix (`cpu_usage_high_v2`, `low_memory_v2`, ...). I guess we have to add the suffix here as well, so that the alerts don't clash with the old rules?

mtyaka (approved):
Nice work @karimelghazouly! :+1:

- [x] I tested this: I ran the prometheus playbook on the server to install the new rules and verified that they get picked up.
- [x] I read through the code

PR #264, opened 2022-02-11
fix: ensure postgres volume condition is correct

gabor-boros:
## Description

The condition for checking if the `POSTGRES_OPENSTACK_DB_DEVICE` is not working with ansible `2.8.17` as expected. We need to use `when: POSTGRES_OPENSTACK_DB_DEVICE != None` instead of `when: POSTGRES_OPENSTACK_DB_DEVICE`.

### DIFF
--- a/playbooks/roles/postgres/tasks/main.yml
+++ b/playbooks/roles/postgres/tasks/main.yml
@@ -39,7 +39,7 @@
     src: "{{ POSTGRES_OPENSTACK_DB_DEVICE }}"
     fstype: ext4
     state: mounted
-  when: POSTGRES_OPENSTACK_DB_DEVICE
+  when: POSTGRES_OPENSTACK_DB_DEVICE != None
 
 - name: chown & chmod /var/lib/postgresql
   file:


playbooks/roles/postgres/tasks/main.yml
```diff
@@ -39,7 +39,7 @@
     src: "{{ POSTGRES_OPENSTACK_DB_DEVICE }}"
     fstype: ext4
     state: mounted
-  when: POSTGRES_OPENSTACK_DB_DEVICE
+  when: POSTGRES_OPENSTACK_DB_DEVICE != None
```
mtyaka, line 42: Nit: I see that in other places in this repo we use `... is not none` instead of `... != None`. It might be worth changing to that for consistency.

mtyaka (approved):
Looks good to me @gabor-boros 👍

playbooks/roles/postgres/tasks/main.yml
gabor-boros: How about a follow up ticket to change all places? I would like to keep the context of the related ticket.

mtyaka: Sure, it's not important, pure cosmetics.

PR #266, opened 2022-02-17
[BB-5416] Mailman patches

Agrendalath:
This adds a note about patching "unknown encoding" errors in Mailman.

Jira ticket: [BB-5416](https://tasks.opencraft.com/browse/BB-5416)


itsjeyd (commented):
Thanks for this @Agrendalath! I'll leave formal approval to @farhaanbukhsh as reviewer of BB-5416, but this is good to go from my point of view. (I read through the text but didn't do any testing beyond checking that alerts subsided after you applied the patch on the existing server.)

farhaanbukhsh (approved):
👍 

- ✅  I tested this, this is already deployed with the patch, and it is working
- ✅  I read through the code
- ✅  Includes documentation

PR #269, opened 2022-03-18
[SE-5407] fix: adds CORS headers in nginx for 500 and 413 response

DubeySandeep:
### Description
413 and 500 response code on `console` server raises CORS issue because of the following reasons:
    - 413: The nginx server doesn't allow request body size < 1 MB and returns 413 response code without any cors header.
    - 500: CORS headers in Django server are handled by `django-cors-headers` middleware but Django structure doesn't allow middleware to attach correct header during 500 errors.
    
 Through this PR we are adding changes to allow Nginx to add correct header for such response code in case the request origin matches the expected pattern.
  
**Ticket link:** https://tasks.opencraft.com/browse/SE-5407

### Testing instructions

__Looking into steps to test this on stage server.__

### DIFF
--- a/playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
+++ b/playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
@@ -13,6 +13,24 @@ server {
 
     {{ NGINX_PROXY_EXTRA_CONFIG }}
 
+    set $expected_origin false;
+    if ($http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) {
+    set $expected_origin true;
+    }
+
+    {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES %}
+    location @{{ response_code }}_cors_header {
+        if ($expected_origin = true) {
+            add_header 'Access-Control-Allow-Origin' $http_origin always;
+            add_header 'Access-Control-Allow-Methods' 'GET,POST,DELETE,OPTIONS' always;
+            add_header 'Access-Control-Allow-Headers' 'Authorization,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range' always;
+        }
+
+        return {{ response_code }};
+    }
+    {% endfor %}
+
     location / {
         proxy_pass {{ NGINX_PROXY_INTERNAL_SCHEME }}://{{ NGINX_PROXY_INTERNAL_DOMAIN }}:{{ NGINX_PROXY_INTERNAL_PORT }}/;
         proxy_http_version 1.1;
@@ -35,5 +53,9 @@ server {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
         {% endif %}
+
+        {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES %}
+        error_page {{ response_code }} = @{{ response_code }}_cors_header;
+        {% endfor %}
     }
 }
--- a/playbooks/roles/ocim/meta/main.yml
+++ b/playbooks/roles/ocim/meta/main.yml
@@ -10,6 +10,8 @@ dependencies:
     NGINX_PROXY_SERVER_NAME: "{{ OPENCRAFT_DOMAIN_NAME }}"
     NGINX_PROXY_SEND_TIMEOUT: 600s
     NGINX_PROXY_READ_TIMEOUT: 600s
+    NGINX_PROXY_ORIGIN_REGEX: '^(http|https)://.*\.opencraft\.com'
+    NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES: [413, 500]
     NGINX_PROXY_EXTRA_CONFIG: "{{ opencraft_nginx_static_config + opencraft_nginx_websocket_config }}"
     when: www_user != 'vagrant'
 


playbooks/roles/ocim/meta/main.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
     NGINX_PROXY_SERVER_NAME: "{{ OPENCRAFT_DOMAIN_NAME }}"
     NGINX_PROXY_SEND_TIMEOUT: 600s
     NGINX_PROXY_READ_TIMEOUT: 600s
+    NGINX_PROXY_ORIGIN_REGEX: '^(http|https)://.*\.opencraft\.com'
```
DubeySandeep, line 13: Should I move this constant to https://github.com/open-craft/ansible-playbooks/blob/777edcf4cd224e446c0a631318bd48b684598711/playbooks/roles/ocim/defaults/main.yml ?

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -13,6 +13,24 @@ server {
 
     {{ NGINX_PROXY_EXTRA_CONFIG }}
 
+    set $expected_origin false;
+    if ($http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) {
+    set $expected_origin true;
```
DubeySandeep, line 18: This is not indented correctly so that the generated config is indented correctly. (following the pattern on the above lines)

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -35,5 +53,9 @@ server {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
         {% endif %}
+
```
DubeySandeep, line 56: Should I leave a code comment explaining why we are handling these cases differently?

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
mtyaka: > This is not indented correctly so that the generated config is indented correctly. (

That's true for lines 21-27, but line not for line 18 above, because that `if` clause is part of nginx config and not jinja code.

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -13,6 +13,24 @@ server {
...
+    {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES %}
+    location @{{ response_code }}_cors_header {
+        if ($expected_origin = true) {
+            add_header 'Access-Control-Allow-Origin' $http_origin always;
+            add_header 'Access-Control-Allow-Methods' 'GET,POST,DELETE,OPTIONS' always;
+            add_header 'Access-Control-Allow-Headers' 'Authorization,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range' always;
+        }
+
+        return {{ response_code }};
```
mtyaka, line 30: Hmm... I guess that means that for 500 errors nginx will discard the response body. That could be annoying when trying to debug errors. I'm starting to think it might be better to change `NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES` to only contain `413` for now.

playbooks/roles/ocim/meta/main.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
...
+    NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES: [413, 500]
```
mtyaka, line 14: Let's remove `500` from this list for now since I think having 500 responses with no CORS headers is preferable to 500 responses with empty request bodies.

mtyaka (changes requested):
This is looking good @DubeySandeep. I added a request to remove 500 from the list of error codes and add a comment to explain why we need a dedicated handler for 413 errors.

In the meantime, I'll set up ansible-playbooks locally to be able to test this.

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -13,6 +13,24 @@ server {
 
     {{ NGINX_PROXY_EXTRA_CONFIG }}
 
+    set $expected_origin false;
+    if ($http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) {
+    set $expected_origin true;
+    }
+
+    {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES %}
```
mtyaka, line 21:
```suggestion
    {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES -%}
```

This adds whitespace control tag to the closing bracket so that the indentation doesn't get messed up.

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -13,6 +13,24 @@ server {
...
+        if ($expected_origin = true) {
+            add_header 'Access-Control-Allow-Origin' $http_origin always;
+            add_header 'Access-Control-Allow-Methods' 'GET,POST,DELETE,OPTIONS' always;
+            add_header 'Access-Control-Allow-Headers' 'Authorization,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range' always;
+        }
+
+        return {{ response_code }};
+    }
+    {% endfor %}
```
mtyaka, line 32:
```suggestion
    {%- endfor %}
```

(for whitespace control)

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -35,5 +53,9 @@ server {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
         {% endif %}
+
+        {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES %}
```
mtyaka, line 57:
```suggestion
        {% for response_code in NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES -%}
```

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -35,5 +53,9 @@ server {
...
+        error_page {{ response_code }} = @{{ response_code }}_cors_header;
+        {% endfor %}
```
mtyaka, line 59:
```suggestion
        {%- endfor %}
```

mtyaka (commented):
I ran ansible and there was no difference in the output other than some whitespace 👍

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
DubeySandeep: Done!

DubeySandeep:
@mtyaka I've addressed the review comments, PTAL!

DubeySandeep:
> I ran ansible and there was no difference in the output other than some whitespace

@mtyaka You mean it generates the correct/expected config file with unwanted extra whitespace OR it's not adding any new config at all?

mtyaka:
@DubeySandeep
>You mean it generates the correct/expected config file with unwanted extra whitespace OR it's not adding any new config at all?

It generates the expected config. Since you already manually installed the changes on stage, running the ansible playbook didn't result in any changes (other than some white-space), which was expected.

DubeySandeep:
> It generates the expected config. Since you already manually installed the changes on stage, running the ansible playbook didn't result in any changes (other than some white-space), which was expected.

Ah, I see! Considering the updated config on Nginx was done manually and contains a handler for 500 issues, I've flushed the changes now, can you please regenerate it and I can help test it on the stage server.

mtyaka (approved):
:+1:

- [x] I tested this: I deployed this to stage and tested that 413 responses contain correct CORS headers.
- [x] I read through the code

PR #270, opened 2022-03-22
[BB-5999] fix: ocim deployment failing due to undefined variable

kaustavb12:
## Description

Deployment of instances in OCIM was failing due to the variable NGINX_PROXY_ORIGIN_REGEX used for nginx configurations was not defined.

## Supporting information

[BB-5999](https://tasks.opencraft.com/browse/BB-5999)

## Testing instructions

1. New [appserver](https://manage.opencraft.com/instance/32680/edx-appserver/22909/) is being setup with this commit version
3. Check deployment proceeds without the error


kaustavb12:
@mtyaka I have added a small check to see if the NGINX_PROXY_ORIGIN_REGEX variable is defined, so as to quickly fix the issue in OCIM.

To actually set the variable for OCIM instances we should look using the [common_configuration_settings](https://github.com/open-craft/opencraft/blob/a0fbf14f9758e7e87037c669af09d08a34c509d6/instance/models/mixins/ansible.py#L69) and set the default value [here](https://github.com/open-craft/opencraft/blob/a0fbf14f9758e7e87037c669af09d08a34c509d6/instance/models/mixins/openedx_config.py#L551-L558) ?

If you think this the correct way to go, then I will go ahead and schedule a task for next sprint to do these changes.

playbooks/roles/nginx-proxy/templates/nginx-proxy.j2
```diff
@@ -14,7 +14,7 @@ server {
     {{ NGINX_PROXY_EXTRA_CONFIG }}
 
     set $expected_origin false;
-    if ($http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) {
+    if (NGINX_PROXY_ORIGIN_REGEX and $http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) {
```
mtyaka, line 17: @kaustavb12 Isn't this still going to fail if the variable is undefined?

I'm guessing a safer approach is to set a default for `NGINX_PROXY_ORIGIN_REGEX` to `null` `nginx-proxy/defaults/main.yml` and then wrap the entire `if ($http_origin ~ {{ NGINX_PROXY_ORIGIN_REGEX }}) { ...` block into a jinja if `{% if NGINX_PROXY_ORIGIN_REGEX %} ... {% endif %}`.

There's another variable which was introduced in https://github.com/open-craft/ansible-playbooks/pull/269 and which I think should also have a default in `nginx-proxy/defaults/main.yml`. I think `[413]` would be a good default.

kaustavb12: @mtyaka You are absolutely right .. in a hurry I completely overlooked the fact that it is jija2 format and that the variable is actually undefined and not null. Thanks a lot for pointing this out.
I have also added the default for the other variable NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES

toxinu:
You may also need to check if `NGINX_PROXY_CORS_HEADERS_FOR_RESPONSE_CODES` is defined.

kaustavb12:
@toxinu I have defined the default for that [here](https://github.com/open-craft/ansible-playbooks/blob/c492fa47c51c1a2a2d22bcfbc782322398772a0d/playbooks/roles/nginx-proxy/defaults/main.yml#L34), so that should not through an error

mtyaka:
Thanks for this @kaustavb12! :+1:

- [x] I tested this: I verified that the test appserver spun up successfully and is up and running.
- [x] I read through the code

toxinu:
> @toxinu I have defined the default for that [here](https://github.com/open-craft/ansible-playbooks/blob/c492fa47c51c1a2a2d22bcfbc782322398772a0d/playbooks/roles/nginx-proxy/defaults/main.yml#L34), so that should not through an error

Perfect! Thanks for fixing this issue @kaustavb12 🙌

PR #273, opened 2022-04-06
Increase tmpreaper runtime for ocim

mtyaka:
Ocim uses many temporary files. Increase tmpreaper runtime since the default often isn't enough for cleaning up temporary files on Ocim.

**Test instructions**:
1. I already ran ansible-playbook with this change on ocim stage & prod.
2. Verify that `/etc/tmpreaprer.conf` contains the new timeout value.

Time will tell if 180 seconds is enough.

### DIFF
--- a/playbooks/group_vars/ocim/public.yml
+++ b/playbooks/group_vars/ocim/public.yml
@@ -4,6 +4,10 @@
 
 COMMON_SERVER_NO_BACKUPS: true
 
+# Ocim uses many temporary files.
+# Increase tmpreaper runtime since the default often isn't enough.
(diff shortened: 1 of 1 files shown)


gabor-boros:
👍  🎉 

- [x] I read through the code

The config is:
```
# Ansible managed
# docs: `man 5 tmpreaper` or https://manpages.ubuntu.com/manpages/focal/man5/tmpreaper.conf.5.html

TMPREAPER_PROTECT_EXTRA=''
TMPREAPER_DIRS='/tmp/.'
TMPREAPER_DELAY='256'
TMPREAPER_ADDITIONALOPTIONS='--runtime=180'
```

PR #275, opened 2022-04-07
[BB-6098] fix: Whitelisting certbot urls for certificate renewal

farhaanbukhsh:
The certbot auto-renewal script was throwing errors at renewal, on investigation we found that certbot looks for a ``well-known/acme-challenge/`` directory which was not present in the webroot so it was throwing a 404 error. On adding  the following lines in ngnix config we set the ``root`` for the URL.

**JIRA tickets**: [BB-6098](https://tasks.opencraft.com/browse/BB-6098)

~~**Discussions**: Link to any public dicussions about this PR or the design/architecture. Otherwise omit this.~~

**Dependencies**: None

~~**Screenshots**: Always include screenshots if there is any change to the UI.~~

~~**Sandbox URL**: TBD - sandbox is being provisioned.~~

~~**Merge deadline**: "None" if there's no rush, "ASAP" if it's critical, or provide a specific date if there is one.~~

**Testing instructions**:

1. ssh in  ``vm-stage.opencraft.com``
2. sudo su and run ``/snap/bin/certbot renew --dry-run`
3. We should see the all successful certificate regenerating properly
4. If we do `vim /etc/nginx/sites-available/default-stage.opencraft.com.conf`
5. We can see the 
```
       location ^~ /.well-known/acme-challenge/ {
                default_type "text/plain";
                root /var/www/certbot;
        }
```
6. If we remove the above lines  and run ``/snap/bin/certbot renew --dry-run` the renew will fail

I have manually added these lines to all 3 config on vm-stage to test 

**Reviewers**
- [ ] @mtyaka

### DIFF
--- a/playbooks/roles/wordpress/templates/default-nginx.j2
+++ b/playbooks/roles/wordpress/templates/default-nginx.j2
@@ -14,18 +14,24 @@ server {
         ssl_certificate /etc/letsencrypt/live/{{ LANDING_SITE_HOSTNAME }}/fullchain.pem;
         ssl_certificate_key /etc/letsencrypt/live/{{ LANDING_SITE_HOSTNAME }}/privkey.pem;
 
-	root /var/www/{{ DEFAULT_SITE_HOSTNAME }};
-	index index.html index.php;
+        root /var/www/{{ DEFAULT_SITE_HOSTNAME }};
+        index index.html index.php;
 
         access_log  /var/www/log/{{ DEFAULT_SITE_HOSTNAME }}_nginx_access.log;
         error_log /var/www/log/{{ DEFAULT_SITE_HOSTNAME }}_nginx_error.log error;
 
-	location / {
-		try_files $uri $uri/ =404;
-	}
+        location / {
+                try_files $uri $uri/ =404;
+        }
 
         # deny access to .htaccess files
         location ~ /\.ht {
                 deny  all;
+        }
+
+        # Certbot renewal challenge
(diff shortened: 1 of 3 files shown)


mtyaka:
Thanks for explaining what the problem is and providing steps to debug @farhaanbukhsh!

I think there's an easier way to solve this: we can remove the plain http (not https) server blocks from the three wordpress nginx configuration files, such as [this one](https://github.com/open-craft/ansible-playbooks/blob/845f93af5a11d8787fb02d56865eec52dc742148/playbooks/roles/wordpress/templates/default-nginx.j2#L1-L6), so that the `default_server` directive from the [certbot nginx config file](https://github.com/open-craft/ansible-playbooks/blob/cc28a0017844dd3eb2e6e6dc107d262a66faebce/playbooks/roles/certbot/templates/nginx-default-server.j2) takes care of the ACME challenge as well as the HTTP -> HTTPS redirect.

I tried commenting out the top `server { ... }` block and the `location ^~ /.well-known/acme-challenge/ { ... }` block from the `/etc/nginx/sites-enabled/default-stage.opencraft.com.conf` file on `vm-stage`, and `/snap/bin/certbot renew --dry-run` worked fine after reloading nginx.

mtyaka (approved):
Thanks for the updates @farhaanbukhsh! :+1:

- [x] I tested this: I applied the update to vm-stage and verified that certbot is able to renew the certificates.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~
- [ ] ~~Includes documentation~~

PR #277, opened 2022-04-18
Postfix now allows smtp servers without login

keithgg:
## Description

Fixes the content declaration of the Postfix password file so
that it's not invalid if not username or password is defined
so that Google's SMTP relay can be used.

The changes in the PR have already been deployed.

## Supporting information

- https://tasks.opencraft.com/browse/SE-5436
- https://github.com/open-craft/ansible-secrets/pull/368

## Testing instructions

- Check out the `ansible-secrets` repository
- Run the playbook with `ansible-playbook -vvv deploy/playbooks/relay.yml --private-key=~/.ssh/vm.pem -l relay.net.opencraft.hosting.yml`
- Mail should work as normal.

## Deadline

18 April 2022

### DIFF
--- a/playbooks/roles/postfix/tasks/main.yml
+++ b/playbooks/roles/postfix/tasks/main.yml
@@ -41,7 +41,7 @@
 - name: configure SMTP client password
   copy:
     dest: "{{ POSTFIX_SASL_CLIENT_PASSWORD_FILE }}"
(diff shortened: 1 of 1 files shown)


mtyaka (approved):
:+1:

- [x] I tested this: I verified that the file on the VM was updated and that mail gets routed correctly.
- [x] I read through the code

PR #284, opened 2022-05-26
Tweak backend response time alert conditions.

mtyaka:
We are getting many false alerts about average response time being too high.
With this tweak we ignore response times  on instances that have very few requests in the past 10 minutes.

The problem was that if an instance would only get two requests in the last 5 minutes for example, and one of the requests took around 2 seconds, the average response time was already above 1 second.

This tries to avoid alerts like these by ignoring instances that received less than 50 requests in the past 10 minutes.

**Testing instructions**:

- Look at the `haproxy_backend_response_time_high` on the [prometheus alerts page](https://prometheus.net.opencraft.hosting/alerts).
- Verify that we've been getting less `haproxy_backend_response_time_high` alerts during the past few days on the ops mailing list.

**Reviewers**:

- [ ] @gabor-boros


gabor-boros (approved):
What a nice tweak! @mtyaka Approved!

- [x] Proofread the changes

mtyaka:
Thanks @gabor-boros!

PR #286, opened 2022-06-08
feat: postfix role now supports multiple relays

keithgg:
## Description
We need to able to route internal emails through Gmail and
client emails through AuthSMTP.

The changes in the commit allow setting a default SMTP provider in
addition to routing to a different provider based on the sender
domain.

The changes of note are:

- Rename `POSTFIX_RELAY_HOST` to `POSTFIX_RELAY_DEFAULT_HOST`
- Removing all the internal relays because they're not needed anymore. `mail.opencraft.com` is only used for the mailing lists.
- Generating the `client-passwd` file based on a list of SMTP details instead of assuming only one is provided.

## Supporting information

- https://tasks.opencraft.com/browse/SE-5564

## Testing instructions

See corresponding secrets PR.

### DIFF
--- a/playbooks/roles/postfix/defaults/main.yml
+++ b/playbooks/roles/postfix/defaults/main.yml
@@ -14,14 +14,22 @@ POSTFIX_TLS_KEY_CONTENT: ""
 # The domain name of mail originating from the mail server (e.g. monitoring emails)
 POSTFIX_MAILNAME: example.com
 
-# The internal domain.  Mail for this domain will be sent to the internal mail server.
-POSTFIX_INTERNAL_DOMAIN: example.com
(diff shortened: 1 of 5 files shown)


mtyaka:
Nice, this simplifies things :+1:

- [x] I tested this: I tested this together with https://github.com/open-craft/ansible-secrets/pull/387
- [x] I read through the code

mtyaka (approved):

mtyaka:
We discovered on [SE-5564](https://tasks.opencraft.com/browse/SE-5564?focusedCommentId=250671&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-250671) that we need to change the configuration to use `inet_protocols = ipv4` instead of `inet_protocols = all` before merging.

PR #287, opened 2022-06-15
feat: use path instead of a different port for Flower in SprinCraft [BB-6393] 

Agrendalath:
This is a part of the cleanup from https://gitlab.com/opencraft/dev/sprintcraft/-/merge_requests/28.

Thanks to this, we can use SSL with Flower.


0x29a (approved):
:+1:

PR #289, opened 2022-06-30
fix: only write postfix certs if there's content

keithgg:
## Description

Fixes the conditional when writing relay certificates so that they don't write anything if there's no content.

## Testing instructions

In related PR.


mtyaka (approved):
It works! :+1:

- [x] I tested this: I tested this together with https://github.com/open-craft/ansible-secrets/pull/396
- [x] I read through the code

PR #291, opened 2022-07-14
chore: remove python3.6 and use system installed python 3.8

keithgg:
## Description

This PR updates cert-manager to run on Ubuntu 20.04 **only**. To achieve this, cert-manager was made to work with Python3.8 in [this PR](https://github.com/open-craft/cert-manager/pull/24).

### Part 1
Run the playbook from the secrets repo.

`ansible-playbook -vvv deploy/playbooks/cert-manager.yml -l cert-manager-stage.net.opencraft.hosting.yml`

There should be no errors.

### Part 2
The changes have already been deployed to `cert-manager-stage.net.opencraft.hosting`. 

- Create a sandbox on https://stage.manage.opencraft.com 
- SSH onto the server
- Verify that your new sandbox's certificate shows up in `/etc/letsencrypt/live/` within about 5 minutes or so after creation.

### DIFF
--- a/playbooks/roles/cert-manager/tasks/main.yml
+++ b/playbooks/roles/cert-manager/tasks/main.yml
@@ -1,21 +1,11 @@
 ---
-
-- name: Add the `deadsnakes` PPA for getting Python3.6 on Ubuntu 16.04
-  apt_repository:
-    repo: 'ppa:deadsnakes/ppa'
-
-- name: Download Python3.6
+- name: Install pip
   apt:
-    name: python3.6
-
-- name: Download pip3.6
-  shell:
-    cmd: curl https://bootstrap.pypa.io/get-pip.py | sudo -H python3.6
+    name: python3-pip
 
 - name: Install pipenv to later install python dependencies and virtual environments.
   pip:
     name: pipenv
-    executable: pip3.6
   become: true
 
 - name: Download cert-manager


playbooks/roles/cert-manager/tasks/main.yml
```diff
@@ -1,27 +1,18 @@
...
   pip:
     name: pipenv
-    executable: pip3.6
   become: true
 
 - name: Download cert-manager
   git:
     repo: "https://github.com/open-craft/cert-manager"
     dest: "{{ cert_manager_path }}"
+    version: "keith/cert-manager-upgrade"
```
keithgg, line 15: If approved, I'll remove this line.

0x29a, line 15: @keithgg, I've followed test steps, but noticed that new directories in `/etc/letsencrypt/live/` have `root` group unlike the existing ones. Is that expected?

0x29a (approved):
:+1:

- [x] I've applied the playbook, created a new instance on the staging and verified that the directory with keys appeared.

playbooks/roles/cert-manager/tasks/main.yml
keithgg: @0x29a Nice find! Yes that's correct. The `ubuntu` group is from my backup. I've rectified it now.

PR #293, opened 2022-08-16
fix: permissions on dovecot directory

keithgg:
## Description

Fixes the permissions on the dovecot directory. When a new server is deployed without this change, there'll be an error message:

```
Aug 16 12:53:08 relay dovecot: auth: Error: passwd-file(discourse,54.38.245.59): stat(/etc/dovecot/private/users) failed: Permission denied (euid=114(dovecot) egid=123(dovecot) missing +x perm: /etc/dovecot/private, dir owned by 0:0 mode=0700)
```
This PR correct the user on the directory so that it can be read.

## Testing instructions

In the secrets repo.


mtyaka:
Nice catch! I tested this and verified that it fixes the permissions issue :+1:

- [x] I tested this: I followed the testing instructions from https://github.com/open-craft/ansible-secrets/pull/405
- [x] I read through the code

PR #295, opened 2022-09-02
fix: header_checks didn't have a handler

keithgg:
## Description
Fixes an issue when deploying the `mail.yml` role. `OPENCRAFT_HEADER_TOKEN` cannot be used in a `when` statement when the value is `null`.

Also adds the handler for the `postmap transport` that was missing.

## Testing instructions

I've added `mail-test.net.opencraft.hosting` in the commit https://github.com/open-craft/ansible-secrets/commit/3feae9ab01d7c85acc6274a86f49cfd35fc75299

Check out the branch `keith/fix-postfix-header-checks` here and in the secrets repo and deploy:

```bash
ansible-playbook -vvv deploy/playbooks/mail.yml -l mail-test.net.opencraft.hosting --private-key=~/.ssh/vm.pem
```

Without these changes you will get an error.

### DIFF
--- a/playbooks/roles/mail/handlers/main.yml
+++ b/playbooks/roles/mail/handlers/main.yml
@@ -36,3 +36,8 @@
   service:
     name: postfix
     state: reloaded
+
+# Update the postfix header checks db
+# This should be run whenever /etc/postfix/header_checks is changed.
+- name: postmap header_checks
+  command: "postmap /etc/postfix/header_checks"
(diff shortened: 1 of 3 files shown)


playbooks/roles/mail/handlers/main.yml
```diff
@@ -36,3 +36,9 @@
   service:
     name: postfix
     state: reloaded
+
+
```
mtyaka, line 40: nit: remove extra newline

```suggestion
```

playbooks/roles/mail/handlers/main.yml
```diff
@@ -36,3 +36,9 @@
...
+# Update the postfix header checks db
+# This should be run whenever /etc/postfix/header_checks is changed.
+- name: postmap transport
```
mtyaka, line 43: I think this should be `postmap header_checks`, the `postmap transport` handler already exists.

```suggestion
- name: postmap header_checks
```

mtyaka (approved):
:+1:

- [x] I tested this: I rand the playbook on the test server, first without the fix (it failed), and then with the fix (it worked!)
- [x] I read through the code

PR #297, opened 2022-09-27
[SE-5766] Remove references to tmpreaper.

mtyaka:
We discovered that tmpreaper doesn't work well together with chkrootkit which (among other things) periodically scans contents of temporary files, thereby updating files' atime values and preventing tmpreaper from being able to purge old files based on their access timestamps.

Since the combination was never working, and we want to keep using chkrootkit, we decided to remove tmpreaper and manually clean up temporary files when disk space gets low.

Note that on the vast majority of our servers/VMs, the amount of temporary files is tiny, with one of the few exceptions being VMs that run Ocim.

**Testing**

1. Checkout this branch and run `ansible-playbook -l ocim-stage deploy/playbooks/ocim.yml --tags=common-server`
2. Verify that the playbook runs successfully, and does not install `tmpreaper` (I manually removed it from Ocim stage).

### DIFF
--- a/playbooks/group_vars/ocim/public.yml
+++ b/playbooks/group_vars/ocim/public.yml
@@ -4,10 +4,6 @@
 
 COMMON_SERVER_NO_BACKUPS: true
 
-# Ocim uses many temporary files.
-# Increase tmpreaper runtime since the default often isn't enough.
-TMPREAPER_RUNTIME: '180'
-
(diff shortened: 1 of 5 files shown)


gabor-boros:
👍  🎉 Approved @mtyaka , nice work!

- [x] I read through the code

gabor-boros (approved):

PR #298, opened 2022-10-10
chore: ssl fixes

keithgg:
## Description

1. Stats page is only available via HAProxy domains
2. Limit vault access to vault domains
   This was not completely possible as `vault` terminates SSL connections itself. This meant that we have to keep the vault port open. However, the change in this PR limits access vault to only `vault` domains.
- Add HSTS headers

## Testing instructions

The changes in this PR have already been deployed to the staging haproxy servers.

WRT to issues mentioned in the description, they can be reproduced by using any domain that had been provisioned on the webserver. For testing purposes, we'll use https://oc-5456.stage.opencraft.hosting

### Stats pages

Visit https://oc-5456.stage.opencraft.hosting:8404, you should get a connection refused error. You should be able to view the stats by going to http://haproxy-stage.net.opencraft.hosting/stats

### Limit vault access

Previously it was possible to go to https://oc-5456.stage.opencraft.hosting:8200 and get back the vault UI. This should no longer be the case (ie. if you're using FF your browser will not use the certificate you loaded).

### Add HSTS headers

These headers are now on any appserver that has been loaded on staging. To test, load any site and view the `Strict-Transport-Security` header.

### DIFF
--- a/playbooks/group_vars/vault/public.yml
+++ b/playbooks/group_vars/vault/public.yml
@@ -11,7 +11,7 @@ vault_tls_dir: "{{ vault_config_dir }}/tls"
 vault_server_scheme: "https"
 
 # Listener stanza
-vault_tcp_address: "0.0.0.0:8200"
+vault_tcp_address: "127.0.0.1:8199"
 vault_tcp_cluster_address: "0.0.0.0:8201"
 vault_tcp_tls_cert_file: "{{ vault_tls_dir }}/cert.pem"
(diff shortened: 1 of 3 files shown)


keithgg:
@gabor-boros When testing I was checking the servers directly rather than going to the parent domain. Turns out the staging domain is `haproxy-stage.opencraft.hosting` (notice the missing `.net`). I've set up DNS for `haproxy-stage.net.opencraft.hosting` which will give you the correct stats page.

gabor-boros:
@keithgg Thank you for the explanation 🙏

PR #299, opened 2022-10-26
fix: removed ssl options from vault cluster

keithgg:
## Description

Fixes the vault cluster communication so that nodes can reach each other.

This PR allows communication through port 8201 to go through without inspecting or changing anything. Previously we were inspecting the SSL headers to determine if the request is being sent to a vault domain. However, that breaks the cluster communication.

Regrettably, this port needs to remain open for the vault cluster to communicate effectively.

## Testing instructions

- Read through the code
- Verify that you can log in on any of the vault staging servers:
  - https://vault-stage-a-1.net.opencraft.hosting:8200/
  - https://vault-stage-b-1.net.opencraft.hosting:8200/
  - https://vault-stage-c-1.net.opencraft.hosting:8200/
  - https://vault-stage.net.opencraft.hosting:8200/

### DIFF
--- a/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
+++ b/playbooks/roles/load-balancer-v2/templates/haproxy.cfg.ctmpl
@@ -96,28 +96,27 @@ backend proxy_stats
 {% if vault_domain_regex %}
 frontend fe_vault
     bind *:{{! external_vault_port !}}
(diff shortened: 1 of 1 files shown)


gabor-boros (approved):
👍  🎉 Nice @keithgg ! Approved.

- [x] I read through the code

PR #301, opened 2022-11-25
refactor: cleanup mattermost resources

gabor-boros:
## Description

This PR removes ansible resources defined for mattermost and mattermail. Since it is already moved to Kubernetes, it is not needed anymore.


keithgg (approved):
@gabor-boros LGTM

- [x] I read through the code

PR #302, opened 2022-12-01
chore: remove RabbitMQ

mtyaka:
This removes everything related to RabbitMQ (we're no longer using it in our infrastructure).

**Testing instructions**:

Verify that there are no traces of rabbitmq left in the repo.


keithgg (approved):
:+1: LGTM

- [x] I read through the code
- [x] I tested this by checking if the string "rabbitmq" still exists in the repository (it doesn't)

PR #308, opened 2023-07-15
Extend hyperkitty lock life

mtyaka:
Extending the lock life so that monthly and hourly jobs don't step on each other's toes.

See:

- https://lists.mailman3.org/archives/list/mailman-users@mailman3.org/thread/HE2LU2KITZCF23UH7K4J2KNMDC65O2N2/
- https://tasks.opencraft.com/browse/SE-6002

**Test instructions**

1. Verify that the setting has been applied to the mail.opencraft.com VM.
2. Wait until the 1st of next month to see if this fixes the issues (unfortunately I cannot think of a better way to test this).


gabor-boros (approved):
Lovely! Approved!

Thank you for the changes, @mtyaka! 🚀

PR #311, opened 2024-07-12
refactor: cleanup after logs.opencraft.com

gabor-boros:
Remove filebeat from the stack as we dismantle logs.opencraft.com and replaced filebeat with fluent-bit.


playbooks/appserver.yml
```diff
@@ -9,10 +9,6 @@
       when: consul_servers | length > 0
     - role: node-exporter
       when: NODE_EXPORTER_PASSWORD is not none
-    - role: filebeat
-      # Appservers may have some Elastic stuff already installed on them.
-      filebeat_elastic_apt_pin: 200
-      when: filebeat_logstash_hosts | length > 0
```
mtyaka, line 15: I don't think we use the `appserver.yml` book anymore (it was only used by Ocim), so we could remove the entire playbook.

playbooks/deploy-all.yml
```diff
@@ -3,7 +3,6 @@
 - import_playbook: accounting.yml
 - import_playbook: backup-pruner.yml
 - import_playbook: elasticsearch.yml
-- import_playbook: elk.yml
```
mtyaka, line 6: We could also remove the elk playbook from this repo if it's no longer needed.

playbooks/appserver.yml
gabor-boros: No, we don't, but I'd like to keep the PR "focused" as this repo will be archived anyways.

mtyaka (approved):
:+1:

- [x] I read through the code

PR #314, opened 2024-10-01
chore: block requests from unknown hosts to accounting application

kaustavb12:
## Description

Adds rules to nginx configs to block requests not coming from know hosts to the accounting app.

## Testing

1. The modified nginx config is deployed in stage.billing.opencraft.com
2. The url `bad-hostname-test.opencraft.hosting` in currently pointing to stage billing server's IP
3. Try hitting this URL, the request should get blocked by nginx

## Other information

[SE-6310](https://tasks.opencraft.com/browse/SE-6310)

### DIFF
--- a/playbooks/roles/accounting/templates/accounting-nginx.j2
+++ b/playbooks/roles/accounting/templates/accounting-nginx.j2
@@ -1,7 +1,7 @@
 server {
     listen 443 ssl;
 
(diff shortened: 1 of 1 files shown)


0x29a (approved):
:+1:

- [x] I tested this: requesting `bad-hostname-test.opencraft.hosting` is blocked by nginx. Also tried this: `curl -H "Host: example.com" https://stage.billing.opencraft.com/admin/`.

